PR #1, opened 2015-10-07
Response

jsmeng324:


lib/assessment.js
```diff
@@ -38,8 +54,17 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
+var addMethod = function addMethod(propertyName, method, obj) {
+  if(typeof arguments[2] === 'undefined') {
+      var newObj = {}
+      newObj[propertyName] = method;
+      return newObj;
+  }
+  else  {
+    obj[propertyName] = method;
+    return obj;
+      }
```
GA-MEB, line 67: This will work, but there's a simpler way. If no value is passed in, `obj` is undefined, so `!obj` would be truthy. Consequently, we could write 

``` javascript
if (!obj) { 
  obj = {};
}
obj[propertyName] = method;
return obj;
```

If we want to be even more terse, we could write that first bit as 

``` javascript
obj = obj || {};
```

since `||` will always evaluate to first truthy value it's given.

GA-MEB:
5 of 5 

Sorry for the hold-up on getting your feedback out.

lib/assessment.js
jsmeng324, line 67: Nice, Matt!
Much cleaner and better code indeed! And...love the || one too! 
I realized the use of arguments[2] was definitely not ideal and could lead to error, but couldn't think of a simpler way to do it at the moment. Need to practice more and read more code for things to connect. :)

PR #2, opened 2015-10-07
completed js-functions-ins-outs assessment

sarabrentmccoy:


lib/assessment.js
```diff
@@ -1,11 +1,19 @@
...
 // Question 1
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
-
+  var minValue = arguments[0];
+  for (var i = 1; i < arguments.length; i++) {
+    if (arguments[i] < minValue) {
+      minValue = arguments[1];
```
GA-MEB, line 13: That `1` should be an `i`, but otherwise this looks right.

Here's the concept. How do we know who the fastest person in the world is? We don't race every potential runner against each other. Instead, we have a record; the fastest person is the one who beats that record.

This principle can be used to find the __est element of any set, be it largest, smallest, or any other single-value score.

lib/assessment.js
```diff
@@ -13,14 +21,23 @@ var min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
-
+  var results = [];
+  for (var i = 0; i < results.length; i++) {
+    total += results[i]
+  }
+  return total;
```
GA-MEB, line 28: This looks right!

lib/assessment.js
```diff
@@ -13,14 +21,23 @@ var min = function min() {
...
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
+// COME BACK
```
GA-MEB, line 34: In case it wasn't clear, in this example, `defaultValue` is a single value, e.g. `"a"`; calling `newArray(5,"a")` would give us back the array `["a","a","a","a","a"].

lib/assessment.js
```diff
@@ -13,14 +21,23 @@ var min = function min() {
...
 var newArray = function newArray(length, defaultValue) {
-
+  var ary = [];
+  for (var i = 0; i < ary.length; i++){
+    ary[] = defaultValue(i);
```
GA-MEB, line 38: Since defaultValue isn't a function, this won't work. 

Additionally, `ary[] =` is not the right syntax. To add something to an array, you can use `.push()`, or you can specify an index and a value, i.e. `ary[3] = "blah"`.

lib/assessment.js
```diff
@@ -29,7 +46,11 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  var ary2 = [];
+  for (var i = 0; i < ary2.length; i++){
+    ary2[] = defaultsFunction(i);
```
GA-MEB, line 51: In this example, `defaultsFunction` is a function, so that part is right. The syntax is still incorrect for adding to an array, though.

lib/assessment.js
```diff
@@ -29,7 +46,11 @@ var newArray = function newArray(length, defaultValue) {
...
+  }
+  return ary2[];
```
GA-MEB, line 53: Also, this should be `return ary2`, since `ary2` is the name of the variable that's holding the array.

lib/assessment.js
```diff
@@ -39,6 +60,12 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
+  if (var obj = {}) {
+    return propertyName.method;
+  };
+  else (
+    return var obj = {};
+  );
```
GA-MEB, line 68: I think this is only answer that's really off-base.
Here's what the code ought to do:
1. Check if `obj` is undefined; if it is, set `obj` to an empty object.
2. Add a new property (with name `propertyName` and value `method`) to `obj`.
3. Return the modified object `obj`.

Translating that to code, we get 

``` javascript
if (!obj) {
 obj = {};
}
obj[propertyName] = method;
return obj;
```

GA-MEB:
3 of 5 (-0.25 on Q1 and Q4, -0.5 on Q3, -1 on Q5)

Nice work. Let me know if you have questions about any of these comments!

PR #3, opened 2015-10-07
Responses to Assessment

baz1389:


lib/assessment.js
```diff
@@ -5,7 +5,13 @@
...
 // assume that all the argument are numbers
 var min = function min() {
-
+  var minimum = arguments[0];
+  for (var i = 0; i < arguments.length; i++){
+    if(minimum < arguments[i]){
+      minimum = arguments[i];
+    }
+  }
+  return minimum; //i think this is right, but can't seem to get it to work
```
GA-MEB, line 14: You've got the sign backwards in line 10 - a number becomes the new minimum by being _less_ than the current minimum. Otherwise, everything is right!

lib/assessment.js
```diff
@@ -29,7 +43,11 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  var ary;
+  for (var i = 0; i < length; i++){
+    ary[i] = defaultsFunction;
+  }
+  return ary; //i think i am unclear on what line 44 is asking - same thing as Q3 but with a function??
```
GA-MEB, line 50: It's saying we should be adding `defaultFunction(i)` to the array each time - invoking `defaultFunction` with the index as the function's input value.

lib/assessment.js
```diff
@@ -40,8 +58,12 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
 
+//i have no idea how to do this on my own
```
GA-MEB, line 61: No worries!
Here's the idea, step by step.
1. Check to see if `obj` has been defined; set it to a blank object to start out with.
2. Add a new method to that `obj` with the name given.

``` javascript
  obj = {};
}
obj[propertyName] = method;
return obj;
```

That first bit could also have been written as `obj = obj || {}`, since `||` will evaluate to the first truthy argument it gets.

lib/assessment.js
```diff
@@ -40,8 +58,12 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
 };
 
+//just FYI, I noticed yesterday that the solution branch was accessible to us (didnt look)
+//It might be beneficial to keep it private until after we take the assessment so we aren't tempted to look at the answers prior to the assessment
+
```
GA-MEB, line 66: We're going by the honor system. People are welcome to look at the answers... it just won't help them (or us) accurately assess their understanding. And at the end of the day, that's the only real point of the assessments anyway.

GA-MEB:
3 of 5 (-0.5 on Q1 and Q4, -1 on Q5)

Nice work. Let me know if you have any questions about my notes on Q5.

PR #5, opened 2015-10-07
Answered questions.

TeoPaolucci:
Had some difficulty with the last one. No references used.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -5,7 +5,18 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
+  if (!arguments[0]) {
+    return undefined;
+  }
 
+  var smallest = arguments[0];
+  for (var i = 1, max = arguments.length; i < max; i++) {
+    if (arguments[i] < smallest) {
+      smallest = arguments[i];
+    }
+  }
+
+  return smallest;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -5,7 +5,18 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
+  if (!arguments[0]) {
+    return undefined;
+  }
```
GA-MEB, line 10: Good thinking, but actually unnecessary in this case; you're setting `smallest` to `arguments[0]` anyway, so even if no arguments are passed in, you're already returning `undefined`

lib/assessment.js
```diff
@@ -13,14 +24,29 @@ var min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
+  if (!arguments[0]) {
+    return 0;
+  }
```
GA-MEB, line 29: See above.

lib/assessment.js
```diff
@@ -39,7 +71,15 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
+  if (!obj) {
+    var target = {};
+  } else {
+    var target = obj;
+  }
+
+  target.prototype.propertyName = method;
```
GA-MEB, line 80: Since we're not using a constructor function to make lots of identical objects, we don't have to worry about making lots of identical copies of our methods; consequently, there's not much to be gained by putting the methods in a separate object (i.e. the constructor function's prototype). 

Additionally, since `target` is not a constructor function, it actually doesn't have a `prototype` property, so that line of code may not run.

lib/assessment.js
```diff
@@ -39,7 +71,15 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
 
+  return target;
```
GA-MEB, line 82: Otherwise, your solution is basically right. A shorter version, however, might be: 

``` javascript
obj = obj || {};
obj[propertyName] = method;
return obj;
```

GA-MEB:
4.5 of 5 (-0.5 on Q5)

Nice work. Let me know if you want to dig deeper into any of these.

PR #6, opened 2015-10-07
Shim-Attempt_1

eshimel:


GA-MEB:
Edie, unfortunately it doesn't look like you got any of these. Take a look through Antony's `solution` branch and see if you can follow them; afterwards, I think it might be a good idea for you to try this one again.

eshimel:
Oh man!  I know!  I totally bombed that one!  I actually went back over it
yesterday to see how I'd do on it days later, and I was pretty close on the
first 3 and still lost on 4 and 5.  I looked over his solutions and I
understood what I doing wrong on those, but 4 and 5 aren't clicking.  Its
on my "get help" list!

PR #7, opened 2015-10-07
responded to questions in assessment.js

gotwilk890:


lib/assessment.js
```diff
@@ -40,6 +89,20 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
+    //creates obj incase one passed through doesn't exist yet
+    obj = {};
+  }
+
+  var newObject = obj;
+
+  //adds method to object
+  newObject[propertyName] = method;
+
+  return newObject;
```
GA-MEB, line 104: All correct. But do you think creating `newObject` is really necessary? Because objects are reference types, `newObject` and `obj` point to the same object; you could just as well have written

``` javascript
obj[propertyName] = method;
return obj;
```

lib/assessment.js
```diff
@@ -40,6 +89,20 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
 
+  //need case where obj is not defined yet
+  if (obj === undefined) {
+
+    //creates obj incase one passed through doesn't exist yet
+    obj = {};
+  }
```
GA-MEB, line 97: Very good! Here's another way to do it:
`obj = obj || {};`

Can you see why that works?

GA-MEB:
5 of 5

PR #8, opened 2015-10-07
Response

gracenjambi:
Used google to check some questions and tried to interpret the code by using different names of variables.


GA-MEB:
Sorry Grace, none of these are correct. Take a look through Antony's solution branch and see if you can follow his answers; afterwards, I think it might be a good idea for you to try this one again.

PR #9, opened 2015-10-07
in out assesment

wiltrahan:
Had a much tougher time with this than I had expected...I thought I was getting somewhere, but when I tried to put it into practice, I went blank.


GA-MEB:
Yeah, unfortunately none of these are right. Take a look through Antony's solution branch and see if you can follow them; afterwards, I think it might be a good idea for you to try this one again.

PR #10, opened 2015-10-07
Finish assessment

dhuddell:
Finish questions
-felt confident 1-3, 4 felt logical, but
question 5 provided some difficulty, wasn't
sure where propertyName and method should go
so I believe I messed up the formation of the function.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -5,7 +5,14 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
-
+  var args = arguments;
+  var min = args[0];
+  for(i=1;i<args.length;i++){
+    if (args[i] < min){
+      min = args[i];
+    }
+  }
+  return min;
 };
 
 // Question 2
@@ -13,14 +20,24 @@ var min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
-
+  var args = arguments;
+  var sum;
+  for(i=0;i<args.length;i++){
+      sum  += args[i];
+    }
+  }
+  return sum;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-
+  var myArray = [];
+  for(i=0;i<length;i++){
+    myArray[i] = defaultValue;
+  }
+  return myArray;
 };
 
 // Question 4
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -5,7 +5,14 @@
...
 var min = function min() {
-
+  var args = arguments;
+  var min = args[0];
+  for(i=1;i<args.length;i++){
+    if (args[i] < min){
+      min = args[i];
+    }
+  }
+  return min;
```
GA-MEB, line 15: Yep, well done.

lib/assessment.js
```diff
@@ -5,7 +5,14 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
-
+  var args = arguments;
```
GA-MEB, line 8: This won't really do anything for you, but it doesn't hurt anything either.

lib/assessment.js
```diff
@@ -29,7 +46,11 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  var myArray = [];
+  for(i=0;i<length;i++){
+    myArray[i] = defaultsFunction(myArray[i]);
```
GA-MEB, line 51: Not quite - the question says that `i` should be the argument, not `myArray[i]`. Additionally, `myArray[i]` doesn't have a value yet, so it doesn't make a ton of sense to pass `undefined` into `defaultsFunction` each time.

lib/assessment.js
```diff
@@ -39,7 +60,14 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
 var addMethod = function addMethod(propertyName, method, obj) {
-
+  this.propertyName = method();     //DAN: confused on where prop and method go.
+  var method = function (obj){
+    if(!obj){
+      return new var newObj{};
+    } else{
+      return obj;
+    }
+  }
```
GA-MEB, line 70: Here's the idea, step by step.
1. Check to see if obj has been defined; set it to a blank object to start out with.
2. Add a new method to that obj with the name given.

``` javascript
  obj = {};
}
obj[propertyName] = method;
return obj;
```

That first bit could also have been written as obj = obj || {}, since || will evaluate to the first truthy argument it gets.

GA-MEB:
3.75 of 5 (-0.25 on Q4, -1 on Q5)

Nice work - let me know if you have any questions about my feedback.

PR #11, opened 2015-10-07
Completed the assessment

AkJones2007:


lib/assessment.js
```diff
@@ -29,7 +50,13 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
+  var array = [];
 
+  for (var i = 0, max = length; i < max; i++) {
+    array.push(defaultsFunction());
```
GA-MEB, line 56: Per the question, that should be `defaultsFunction(i)`.

lib/assessment.js
```diff
@@ -39,9 +66,18 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
+    obj = new NewObject;
+  }
 
+  obj.prototype.propertyName = method;
 };
 
+var result;
+result = newArray2(4, function () {return 'value';});
+console.log(result);
+
```
GA-MEB, line 80: I think you overthought this one.
Here's the idea, step by step.
1. Check to see if `obj` has been defined; set it to a blank object to start out with.
2. Add a new method to that `obj` with the name given.

``` javascript
  obj = {};
}
obj[propertyName] = method;
return obj;
```

That first bit could also have been written as `obj = obj || {}`, since `||` will evaluate to the first truthy argument it gets.

GA-MEB:
3.75 of 5 (-0.25 of Q4, -1 on Q5)

Nice work. Let me know if you have questions on Q5.

lib/assessment.js
```diff
@@ -39,9 +66,18 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
+  if (!obj) {
+    var NewObject = {};
+    obj = new NewObject;
+  }
 
+  obj.prototype.propertyName = method;
```
GA-MEB, line 74: Prototypes only get used when we're working with constructor functions.

PR #12, opened 2015-10-07
Completed assessment

raq929:


lib/assessment.js
```diff
@@ -5,7 +5,16 @@
...
+  var arg = arguments;
+  if(arg){
+    var minimum = arg[0];
+    // this for loop is not working, and I'm not sure why. It's not entering the for loop.
+    for(var i = 0, length = arg.lenth; i < length; i++){
+      if(arg[i] < minimum){
+        minimum = arg[i];
+      }
+    } return minimum;
+  } else return console.log("What is the minimm of nothing?");
```
GA-MEB, line 17: I tested your code, and (with a few minor syntactical corrections) it seems to work. See below:

``` javascript
var min = function min() {
    var arg = arguments;
    if(arg){
        var minimum = arg[0];
        for(var i = 0, length = arg.length; i < length; i++){
          if(arg[i] < minimum){
            minimum = arg[i];
          }
        } 
        return minimum;
    } else {
        return console.log("What is the minimm of nothing?");
    }    
 };
```

lib/assessment.js
```diff
@@ -13,14 +22,24 @@ var min = function min() {
...
 var sum = function sum() {
-
+  var arg = arguments;
+  var total = 0;
+  //not entering the if statement. Sigh.
+  if(arg){
+    for (var i = 0, length = arg.lenth; i < length; i++) {
+      total += arg[i];
+    }
+  } return total;
```
GA-MEB, line 32: Ditto to above.

``` javascript
var sum = function sum() {
    var arg = arguments;
    var total = 0;
    if(arg){
        for (var i = 0, length = arg.length; i < length; i++) {
            total += arg[i];
        } 
        return total;
    }    
};
```

GA-MEB, line 32: Incidentally, it's not really necessary to nest all your code in an if - what would be simpler is just handling special cases up front and returning early.
e.g.

``` javascript
var sum = function sum() {
  if (arguments.length === 0) {
   return 0;
  }
  var total = 0;
   total += arguments[i];
  }
  return total;
};
```

In this case, though, it's actually not even necessary to do something like that, since if there are no arguments, the `for` loop never runs, and so we just return 0 - the code below could be left out entirely.

``` javascript
  if (arguments.length === 0) {
   return 0;
  }
```

lib/assessment.js
```diff
@@ -29,7 +48,10 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  var array = [];
+  for(var i = 0; i < length; i++){
+    array[i] = defaultsFunction(i);
```
GA-MEB, line 53: Yep. Push would also work here.

lib/assessment.js
```diff
@@ -39,7 +61,11 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
-
+  if(!obj){
+    obj = {};
+  }
+  obj[propertyName] = method;
+  return obj;
```
GA-MEB, line 68: VERY nice! Spot on with this one.

Another way that you could have written the first bit is `obj = obj || {};`.
Can you see why?

GA-MEB:
4.5 of 5 (-0.25 on Q1, Q2)

Nice job. Let me know if you have any questions.

PR #13, opened 2015-10-07
finished the assessment

scottyscripts:
didn't have enough time to finish.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -4,23 +4,33 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-var min = function min() {
-
+var min = function min(nums) {
+  return Math.min(nums)
 };
+
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
-
+var sum = function sum(addends) {
+  var adding = 0;
+  for (i = 0; i < addends.length; i++) {
+    adding += addends[i];
+  }
+  return adding;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-
+  ary = [];
+  ary.length = length;
+  for (var i = 0; i < length; i ++) {
+    ary[i] = defaultValue;
+  }
+  return ary;
 };
 
 // Question 4
@@ -29,8 +39,16 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  ary = [];
+  ary.length = length;
+  defaultsFunction = function defaultsFunction(element) {
+    for (var i = 0; i < length; i ++) {
+    ary[i] = element;
+  }
+  }
+  return ary;
 };
+
 
 // Questions 5
 // Write a function that adds or replaces a method on an object
@@ -39,6 +57,8 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
+  function obj
+
 
 };
 


lib/assessment.js
```diff
@@ -4,23 +4,33 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-var min = function min() {
-
+var min = function min(nums) {
+  return Math.min(nums)
```
GA-MEB, line 8: This will not work, because `nums` now refers to only the _first_ argument passed in. 

(it also kinda misses the point, which is to see how you might implement `Math.min` on your own)

lib/assessment.js
```diff
@@ -4,23 +4,33 @@
...
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
-
+var sum = function sum(addends) {
+  var adding = 0;
+  for (i = 0; i < addends.length; i++) {
+    adding += addends[i];
+  }
+  return adding;
```
GA-MEB, line 21: Now **this** looks more like it. There's just one problem, and it's the same as what I mentioned above.

To access the arguments passed into a function, irrespective of how many there are, you need to use the `arguments` array-like object.

If you drop in `argument` instead of `addends` everywhere in the body, and take `addends` out of the list of parameters, the code should work.

lib/assessment.js
```diff
@@ -29,9 +39,17 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  ary = [];
+  ary.length = length;
```
MattB-GA-Boston: This line will not do what you think it does. In fact, don't ever write this, because it will really mess up the array that you do it to. `.length` is a attribute of an array, not a method; the way that it's kept accurate is that every time an element gets added or removed, the `.length` value gets updated automatically to account for the change. If you manually set `.length` to a different value - say 10 - if you add one more element, `.length` will be equal to 11 _irrespective of how long the array actually is_. See why that might be dangerous?

lib/assessment.js
```diff
@@ -29,9 +39,17 @@ var newArray = function newArray(length, defaultValue) {
...
+  defaultsFunction = function defaultsFunction(element) {
```
GA-MEB, line 44: Not sure why this is here. It's not necessary to enclose the `for` loop in a function.

lib/assessment.js
```diff
@@ -29,9 +39,17 @@ var newArray = function newArray(length, defaultValue) {
...
+    for (var i = 0; i < length; i ++) {
+    ary[i] = element;
+  }
+  }
+  return ary;
```
GA-MEB, line 49: Otherwise this solution is correct.

lib/assessment.js
```diff
@@ -39,6 +57,8 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
+  function obj
+
```
GA-MEB, line 61: Here's what we were looking for:

``` javascript
var addMethod = function(propertyName, method, obj) {
  if (!obj) { obj = {}; }  // Alternatively, `obj = obj || {};`. Can you see why?
  obj[propertyName] = method;
  return obj;
};
```

lib/assessment.js
```diff
@@ -4,23 +4,33 @@
...
+  return adding;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-
+  ary = [];
+  ary.length = length;
```
GA-MEB, line 29: This line will not do what you think it does. In fact, don't ever write this, because it will really mess up the array that you do it to. `.length` is a attribute of an array, not a method; the way that it's kept accurate is that every time an element gets added or removed, the `.length` value gets updated automatically to account for the change. If you manually set `.length` to a different value - say 10 - then if you add one more element, `.length` will be equal to 11 _irrespective of how long the array actually is_. See why that might be dangerous?

lib/assessment.js
GA-MEB, line 43: See note above.

lib/assessment.js
```diff
@@ -4,23 +4,33 @@
...
+  }
+  return adding;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-
+  ary = [];
```
GA-MEB, line 28: Should be `var ary`.

lib/assessment.js
```diff
@@ -29,9 +39,17 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  ary = [];
```
GA-MEB, line 42: See note above.

lib/assessment.js
```diff
@@ -29,9 +39,17 @@ var newArray = function newArray(length, defaultValue) {
...
+  ary.length = length;
+  defaultsFunction = function defaultsFunction(element) {
+    for (var i = 0; i < length; i ++) {
+    ary[i] = element;
```
GA-MEB, line 46: Each element in the array should have a value equal to `defaultsFunction(i)`, for index value `i`.

GA-MEB:
1.75 of 5 (-1 on Q1, Q5; -0.5 on Q2, Q4; -0.25 on Q3)

You're close on a lot of these, but `arguments` seems to be giving you some trouble. 
Might be worthwhile to try this one again when you have time.

PR #14, opened 2015-10-07
won't take too long to look over this one...

jcmeyer10:


lib/assessment.js
```diff
@@ -5,15 +5,19 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
+  for (i=0; i< arguments.length; i++){
+    return arguments+
+  }
+min(1,4,9);
```
GA-MEB, line 11: Here's the concept. `arguments` is an array-like object that holds all the arguments our function gets. If we want to get the minimum value of those arguments, we can iterate through the array and 'memo-ize' (store between iterations) our minimum value, comparing each element to the minimum value and updating the minimum value accordingly.
i.e.

``` javascript
var min = function (){
  var minimum = arguments[0];
  for (var i = 1; i < arguments.length; i++) {
    if (arguments[i] < minimum) {
      minimum = arguments[i];
    }
  }
  return minimum;
};
```

lib/assessment.js
```diff
@@ -5,15 +5,19 @@
...
+// I know this is wrong, I need to put args in min(args) but cant remember how to set it to the next variable because when I type in arg = arguments, arugments is already defined and my code will only accept three cases right now because of how its set up as a simple if/then statement.
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
-
+var sum = function sum(arg1,arg2) {
+  return arg1 + arg2;
```
GA-MEB, line 20: Now imagine that the concept from Q1 is applied to Q2. You might get something like this:

``` javascript
var sum = function(){
  var total = 0;
    total += arguments[i];
  }
  return total;
};
```

lib/assessment.js
```diff
@@ -21,6 +25,7 @@ var sum = function sum() {
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
 
+//didn't even get this far, just need to see it again.  couldn't find antony's code along.  can you please update any repo with the code along so if we go over it again, it's not a blank with the starter code.
```
GA-MEB, line 28: Here's one possible solution.

``` javascript
var newArray = function(length, defaultValue) {
  var arr = [];
  for (var i = 0; i < length; i++) {
    arr[i] = defaultValue;
  }
  return arr;
};
```

lib/assessment.js
```diff
@@ -32,6 +37,8 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 
 };
 
+didn't even get this far, just need to see it again.  couldn't find antony's code along.  can you please update any repo with the code along so if we go over it again, it's not a blank with the starter code.  i will use my lunch time or after the day is over to go over this.  just let me know.
+
```
GA-MEB, line 41: Given my comment above, do you think you could come up with a solution to this question now?

lib/assessment.js
```diff
@@ -41,6 +48,7 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 var addMethod = function addMethod(propertyName, method, obj) {
 
 };
+/didn't even get this far, just need to see it again.  couldn't find antony's code along.  can you please update any repo with the code along so if we go over it again, it's not a blank with the starter code.
```
GA-MEB, line 51: Here's a solution to this problem:

``` javascript
var addMethod = function (propertyName, method, obj) {
  if (!obj) {
    obj = {};
  }
  obj[propertName] = method;
  return obj;
};
```

GA-MEB:
Definitely would encourage you to try this one again, if you have time - I'd be happy to give you some more feedback!

PR #15, opened 2015-10-07
my answers to assessment.js for functions-ins-and-outs

christinemoses:


lib/assessment.js
```diff
@@ -4,15 +4,30 @@
...
 // assume that all the argument are numbers
-var sum = function sum() {
+var sum = function sum([1, 2, 3, 4]) {
+
+  var sumTotal = 0;
+  for(var i = 0; i < sum.length; i++); {
+    sumTotal += sum[i];
+  }
+
+  return sumTotal;
```
GA-MEB, line 30: When you write a function that takes an unknown number of arguments, you want to use the `arguments` array-like object which is built into each function. This object allows you to access each argument using an index value, just like an array.

In this case, you would write the following.

``` javascript
var sum = function sum() {
  var sumTotal = 0;
    sumTotal += arguments[i];
  }
  return sumTotal;
};
```

Otherwise, your logic for approaching this problem is spot on.

lib/assessment.js
```diff
@@ -4,15 +4,30 @@
...
 // and returns the minimum from that set
 // assume that all the argument are numbers
-var min = function min() {
+var min = function min([1, 3, 4, 2]) {
+
+  var findMin = 0;
+
+  for(var i = 0; i < min.length; i++) {
+    return Math.min.apply();
+  };
```
GA-MEB, line 13: See my point below (Q2). In this case, one possible solution might be:

``` javascript
var min = function min() {
  var minimum = arguments[0];
  for (var i = 1; i < arguments.length; i++) {
    if ( arguments[i] < minimum ) {
      minimum = arguments[i];
    }
  }
  return minimum;
};
```

GA-MEB:
I'd recommend trying this one again when you get some time - let me know if you'd like feedback on your answers!

PR #16, opened 2015-10-07
assessment almost done

leahphuong:


lib/assessment.js
```diff
@@ -20,9 +38,17 @@ var sum = function sum() {
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
+  //create a new empty array
+  //loop through length and push default Value to the new empty arrays
+    var Array = [];
```
GA-MEB, line 43: As a rule, don't capitalize any variables in JavaScript unless they're holding/referring to Constructor functions.

lib/assessment.js
```diff
@@ -20,9 +38,17 @@ var sum = function sum() {
...
+    for(var i = 0; i < length; i++){
 
+     Array[i] = defaultValue;
+    return Array;
```
GA-MEB, line 47: The `return` statement will need to be outside of the `for` loop in order for this to work correctly.

GA-MEB:
First two are correct, and the third is mostly right. I recommend giving this another attempt sometime later this week, if you have time.

PR #17, opened 2015-10-07
Finished assessment

atsundwn:


lib/assessment.js
```diff
@@ -39,7 +61,13 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
-
+  if(obj[propertyName] === undefined) {
+    obj[propertyName] = method;
+    return obj;
+  } else {
+    obj[propertyName] = method;
+    return obj;
+  }
```
GA-MEB, line 70: Not quite. This doesn't check to see if `obj` is undefined, it checks to see if `obj[propertyName]` is undefined. Still, very close! 

Here's what we were looking for.

``` javascript
if (!obj) {
  obj = {};
} // alternatively, we could have written `obj = obj || {};` - can you see why that works?
obj[propertyName] = method;
return obj;
```

GA-MEB:
4.5 of 5.

Nice job, Min. Let me know if you have any questions about my feedback.

PR #18, opened 2015-10-07
Response

kelvinma:
Referred to class notes for Question 4.

Referenced stack overflow for question 5.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -6,7 +6,7 @@
 // assume that all the argument are numbers
 var min = function min() {
   var sorted = [];
-  for (i = 0; i < arguments.length; i++) {
+  for (var i = 0; i < arguments.length; i++) {
     sorted[i] = arguments[i];
   }
     sorted.sort(function(a,b){
@@ -21,7 +21,7 @@ var min = function min() {
 // assume that all the argument are numbers
 var sum = function sum() {
   var total = 0;
-  for (i = 0; i < arguments.length; i++) {
+  for (var i = 0; i < arguments.length; i++) {
     total += arguments[i];
   }
   return total;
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -5,7 +5,14 @@
...
 var min = function min() {
-
+  var sorted = [];
+  for (var i = 0; i < arguments.length; i++) {
+    sorted[i] = arguments[i];
+  }
+    sorted.sort(function(a,b){
+      return a - b;
+  });
+    return sorted[0];
```
GA-MEB, line 15: Clever! This is a novel way of getting the solution.
Assuming that you didn't have `sort` at your disposal, though, you might have to revert to doing things the hard way, e.g. using a `for` loop.

lib/assessment.js
```diff
@@ -29,7 +44,8 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  var array2 = newArray(length, defaultsFunction);
+  return array2;
```
GA-MEB, line 48: Not quite - the directions say that each element should have the value `defaultFunction(i)`. 
i.e.
`[defaultFunction(0), defaultFunction(1), ...]`
Given that, do you think you could rewrite your solution?

lib/assessment.js
```diff
@@ -39,7 +55,12 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
-
+  if (obj) {
+    this.propertyName = method;
+  } else {
+      var newObj = {};
+      newObj.propertyName = method;
+  }
```
GA-MEB, line 63: Almost! Dot notation won't work here - you'll just end up with a property that's literally called `propertyName`. Instead, you need to use square bracket notation.

GA-MEB:
4 of 5 (-0.5 on Q4, Q5)

Very nice work. Let me know if you have questions about any of my feedback.

PR #20, opened 2015-10-07
Completed assessment

kaylal12:


lib/assessment.js
```diff
@@ -5,31 +5,63 @@
...
+
 // Question 4
 // Write a function that returns an array of the specified length
 // with all the elements set to the value returned from
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
+  var newAry = []
+  for (var i = 0; i < length; i++) {
+    newAry[i] = defaultsFunction();
```
GA-MEB, line 61: Close - that should be `defaultsFunction(i)`

lib/assessment.js
```diff
@@ -39,7 +71,15 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
-
+  // not completely sure how to approach
+  if (!obj) {
+    return {};
+  }
+  else {
+    function(obj) {
+      propertyName.method = propertyName.method;
+    }
+  }
```
GA-MEB, line 82: Clarifying the question may help here. This method is supposed to take an object and add a method to it. But what if no object is passed in? In that case, we should make a new object, and add the method to that.

As a set of steps, that becomes:
1. Check to see if `obj` is undefined or not; if it is, set `obj` to an empty object.
2. Add new property (propertyName: method) to `obj`
3. Return `obj`

As code, that could be written

``` javascript
if (!obj) { obj = {}; } //Alternatively, 'obj = obj || {};' - can you see how that might work?
obj[propertyName] = method;
return obj;
```

GA-MEB:
4 of 5 (-0.25 on Q4, -0.75 on Q5)

Nice work - let me know if you have questions about any of my feedback.

PR #21, opened 2015-10-07
added response to challenge.js

resource11:
Struggled with #5... probably could figure it out with a bit more time. Had trouble deciphering the exact needs in that one question.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -5,7 +5,18 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
+  // get the arguments, assign to args
+  var args = arguments;
 
+  var result = 0;
+
+  // iterate over args length and compare value to result
+  for (var i = 0; i < args.length; i++) {
+    if (result < args[i]) {
+      result = args[i];
+    }
+  }
+  return result;
 };
 
 // Question 2
@@ -13,7 +24,17 @@ var min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
+  // get the arguments, assign to args
+  var args = arguments;
 
+  var result = 0;
+
+  // iterate over args length and compare value to result
+  for (var i = 0; i < args.length; i++) {
+    result += args[i];
+    }
+  }
+  return result;
 };
 
 // Question 3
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -5,7 +5,18 @@
...
 
+  var result = 0;
+
+  // iterate over args length and compare value to result
+  for (var i = 0; i < args.length; i++) {
+    if (result < args[i]) {
+      result = args[i];
+    }
+  }
+  return result;
```
GA-MEB, line 19: Yep!

One small point - using `args` instead of `arguments` doesn't actually give you any advantage. Both variables point to the exact same object.

lib/assessment.js
```diff
@@ -30,6 +62,17 @@ var newArray = function newArray(length, defaultValue) {
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
 
+ var result = [];
+
+  // iterate over the specified length
+  for (var i = 0; i < length; i++) {
+    // set each index value of the result of invoking the defaultsFunction
+    result[i] = defaultsFunction();
```
GA-MEB, line 70: Close - should be `defaultsFunction(i)`.

lib/assessment.js
```diff
@@ -40,6 +83,24 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
+  }
+
+    if (/*obj is unset*/) {
+      var newObj = obj;
+      } else {
+      // return the already-in-existence object;
+    }
+
+  return result;
+
```
GA-MEB, line 103: Clarifying the question may help here. This method is supposed to take an object and add a method to it. But what if no object is passed in? In that case, we should make a new object, and add the method to that.

As a set of steps, that becomes:
1. Check to see if `obj` is undefined or not; if it is, set `obj` to an empty object.
2. Add new property (propertyName: method) to `obj`
3. Return obj

As code, that could be written

``` javascript
if (!obj) { obj = {}; } //Alternatively, 'obj = obj || {};' - can you see how that might work?
obj[propertyName] = method;
return obj;
```

GA-MEB:
3.75 of 5 (-0.25 on Q4, -1 on Q5)

Nice work! Let me know if you'd like to chat more about any of these questions.

lib/assessment.js
resource11, line 19: Ah, OK. I thought we received explicit instruction to not use 'arguments' without assigning that to a variable.

lib/assessment.js
resource11, line 70: Ahh, yes. Thanks!

PR #23, opened 2015-10-07
assesment

drewster317:


lib/assessment.js
```diff
@@ -4,24 +4,37 @@
...
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
+  var data = [];
+
+  for(var i = 0; i < length; i++) {
+      data[i] = defaultValue;
+
+  }
+
+return data;
```
GA-MEB, line 34: Nice. This solution looks good.

lib/assessment.js
```diff
@@ -4,24 +4,37 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-var min = function min() {
+var min = function min(num1,num2) {
+  return Math.min(num1,num2);
```
GA-MEB, line 8: Because we don't know how many arguments our function will be given, we have to use the array-like object `arguments` to keep track of whatever arguments we get. 

Additionally, the idea was more that you would create the functionality of `Math.min` from scratch. Can you think about how you might do that? 
(HINT: It's doable with a single `for` loop.)

lib/assessment.js
```diff
@@ -4,24 +4,37 @@
...
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
+var sum = function sum(num1,num2,num3) {
+  return num1 + num2 + num3
 
 };
+console.log(sum(1,2,3));
```
GA-MEB, line 21: See my note above.

lib/assessment.js
```diff
@@ -29,8 +42,19 @@ var newArray = function newArray(length, defaultValue) {
...
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
+  var data = [];
+
+  for(var i = 0; i < length; i++) {
+      data[i] = defaultsFunction(i);
+
+  }
+
+return data;
```
GA-MEB, line 52: Exactly right!

GA-MEB:
2.5 of 5 (-0.75 on Q1 and Q2, -1 on Q5)

Nice work, particular on Q3 and Q4. Let me know if you have questions, or ever want to re-take this assessment - I'd be happy to give you feedback again.

PR #24, opened 2015-10-07
need more time

JT0Y:


lib/assessment.js
```diff
@@ -4,32 +4,57 @@
...
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
-
+    var total=0;
+    var i = 0;
+    while(i < arguments.length) {
+        total =+ arguments[i] + i + 1;
+        i++;
+    }
```
GA-MEB, line 32: I would probably have suggested using a `for` loop here, but using a `while` is ok. 

However, it should be `+=`, not `=+`. Additionally, it's not clear why you're adding `+ i + 1` each time you go through the `while` loop.

lib/assessment.js
```diff
@@ -4,32 +4,57 @@
...
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-
+    i = 0;
+    array=[]
+    while (i < defaultValue){
+        array.push(length);
```
GA-MEB, line 43: Why are you pushing the length? If we call `newArray(5,'a')`, we want an array of five `'a'`s, like so:
`['a','a','a','a','a']`.
It's `defaultValue` that should be pushed to the array.

GA-MEB:
Definitely take a look through the `solution` branch for this assessment, and see if you can follow along - if it all makes sense to you, I recommend taking this again. I'd be happy to give you feedback if you submit another PR.

PR #26, opened 2015-10-07
got to attempt the first three

francescadal:
difficult


srhoulam:
Check this pull request to see that what you submitted is actually all of your work. All I see is a `return` statement in the first two functions. Please resubmit this as another pull request if it was a mistake.

0/5

PR #27, opened 2015-10-07
Answered assessment

HoldenMills:


lib/assessment.js
```diff
@@ -4,24 +4,31 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-var min = function min() {
-
+var min = function min(arg) {
+  min = math.min();
```
srhoulam, line 8: A few notes:
- Assigning to `min` like this will replace the reference to this function with whatever is on the right of the equal sign. The first time you run it, it will overwrite the `min` variable. The next time you invoke it, `min` will refer to something else.
- Identifiers (e.g., variable names) are case-sensitive. `Math` and `math` are different variables.
- This function was intended to have an empty parameter list. This lesson introduced the `arguments` pseudo-array, available only inside functions, which contains the argument values passed to the function when it was invoked.
- Finally, using the built in methods on the `Math` object is cheating! You were supposed to write it yourself!

lib/assessment.js
```diff
@@ -4,24 +4,31 @@
...
+//I used mdn
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
-
+  return math.sum();
 };
```
srhoulam, line 18: See above.

Here, you try to return the _invoked_ method, which means it returns some other value. `math` is not `Math`, and `Math.sum` doesn't exist anyway.

lib/assessment.js
```diff
@@ -4,24 +4,31 @@
...
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
+  var newArray = [defaultValue];
```
srhoulam, line 24: If the `for` loop were correct, this wouldn't matter, but it's important to understand. This does not create an empty array, but a one-element array containing the value stored in `defaultValue`.

lib/assessment.js
```diff
@@ -4,24 +4,31 @@
...
+  for (i=0; i < length.length; i++) {
+    return newArray[i];
```
srhoulam, line 26: When a `return` statement is encountered in a function, it stops running and returns to where it was called from with the specified value.

What happens here is, rather than looping, it does the first iteration (where `i` is zero) and returns with the 0th element of `newArray`. Ultimately, this function is returning `defaultValue`.

lib/assessment.js
```diff
@@ -4,24 +4,31 @@
...
-
+  return math.sum();
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
+  var newArray = [defaultValue];
+  for (i=0; i < length.length; i++) {
```
srhoulam, line 25: Did you mean `newArray.length`? Arrays have a `length` property. Numbers do not.

lib/assessment.js
srhoulam, line 24: Finally, on naming. Naming this variable `newArray` works. The challenge is that it denies you access to another variable called `newArray` that points to the function that is currently executing. It won't throw an error when you do it here, but you should be conscious that you are making your function incapable of referencing itself when you create a variable of the same name. However, the greatest damage is done to readability. Someone else reading your code (including yourself in the future) will find this confusing.

srhoulam:
0/5

PR #28, opened 2015-10-07
assessment

pguzman26:


lib/assessment.js
```diff
@@ -13,6 +14,9 @@ var min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
+  var one = number(one);
+  var two = number(two);
+  return number(one+two);
 
 };
```
srhoulam, line 21: You are not provided with a function called `number`. You were expected to loop over the `arguments` pseudo-array within the function to get the values passed in and add them up.

lib/assessment.js
```diff
@@ -20,7 +24,10 @@ var sum = function sum() {
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
+  var length = this.length;
```
srhoulam, line 27: Since this function is not called as a method, `this` will resolve either to the global object or `undefined`.

lib/assessment.js
```diff
@@ -20,7 +24,10 @@ var sum = function sum() {
...
+
 
+return newArray;
 };
```
srhoulam, line 31: Since `newArray` is the variable this function is assigned to, and also the name for the function within itself, this function returns itself. Functions that return themselves are, in most cases, not what you want to create.

srhoulam:
0/5

PR #29, opened 2015-10-07
updated response file

bbarbersox:


lib/assessment.js
```diff
@@ -5,7 +5,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
-
+  lowestNumber = argument[0];
```
srhoulam, line 8: `arguments`, not `argument`. This is clearly just a typo.

lib/assessment.js
```diff
@@ -5,7 +5,13 @@
...
+  for (var i=1; i < arguments.length; i++) {
+    if (arguments[i] < lowestNumber) {
+      lowestNumber = arguments[i];
+    }
+  }
+  return lowestNumber;
 };
```
srhoulam, line 15: Great. This does exactly what's requested, aside from the typo.

lib/assessment.js
```diff
@@ -13,14 +19,25 @@ var min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
-
+  var total = 0;
+  for (var i = 0; i < arguments.length; i++) {
+    total = total + arguments[i];
+  }
+  return total;
 };
```
srhoulam, line 27: Yep. Keep it up.

lib/assessment.js
```diff
@@ -13,14 +19,25 @@ var min = function min() {
...
 var newArray = function newArray(length, defaultValue) {
-
+var i = 0;
+while (length > 0) {
+  defaultArray[i] = defaultValue;
+  length = length - 1;
+  i = i + 1;
+}
+return defaultArray;
 };
```
srhoulam, line 41: So, this function would have been great if the array it returns were created inside the function. Since it's created outside the function, and referenced inside it as an external variable, every time this function is run, all it does is overwrite the same array. That means _the_ (as in it returns the same specific array each time) array returned by this function is **extremely unsafe!** It can be overwritten unexpectedly without notice! Half credit.

lib/assessment.js
```diff
@@ -13,14 +19,25 @@ var min = function min() {
...
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
+var defaultArray = [];
 var newArray = function newArray(length, defaultValue) {
-
+var i = 0;
+while (length > 0) {
+  defaultArray[i] = defaultValue;
+  length = length - 1;
```
srhoulam, line 37: We usually don't want to change the values of the parameters we take into the function unless that is their express purpose. Using it as a loop countdown as you did here is a destructive operation, meaning we lose the original state of the parameter and can't reconstruct it afterward without another piece of information.

This can be resolved by changing the loop condition.

lib/assessment.js
```diff
@@ -13,14 +19,25 @@ var min = function min() {
...
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
+var defaultArray = [];
 var newArray = function newArray(length, defaultValue) {
-
+var i = 0;
+while (length > 0) {
```
srhoulam, line 35: By choosing to test for `length > 0`, you have constrained yourself into using `length` as a countdown for this loop. We'd rather not alter our parameters' values, as we may need them later. Instead, consider counting _up_ from `i` to `length` by setting the loop condition to `i < length`.

lib/assessment.js
```diff
@@ -29,7 +46,11 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  for (var i = 0; i < length; i++) {
+    newArray2 [i] = defaultsFunction();
+    // newArray2 [i] =
+  }
+  return newArray2;
 };
```
srhoulam, line 54: Here, we see a similar problem to above, in that you did not declare a new variable, create a new array, and assign the array to the variable. Imagine this is the first line in your function's body, before the `for` loop: `var arrayToReturn = [];`

Without an array to work with, and a for loop to put a statement inside, anyone would be confused about what to do on line 50. What your line 50 does is assign the result of invoking `defaultsFunction` without any arguments to a numeric property on the function that is currently executing. Finally, the function returns itself.

The reason I describe this as similar to your answer to question 4 is because it returns the same reference every time you run it. If you were to invoke this function several times, the function object (stored in `newArray2`) will have attributes added to it and overwritten in subsequent invocations, and the return value would always be the function itself. Anything accessing the attributes of the returned function will be working with an unsafe store of data that can be overwritten without notice at any time.

1/4 credit.

srhoulam:
2.75/5

PR #31, opened 2015-10-12
Original PR was from the wrong assessment branch. This is the corrected version.

LaurenGrant:


lib/assessment.js
```diff
@@ -5,7 +5,15 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
+  var minimum = arguments[0];
+  for (var i = 0; i < arguments.length; i++) {
+    if (arguments[1] < minimum) {
```
srhoulam, line 10: This tests whether the 1th index of `arguments` is less than `minimum` each iteration of the loop. This leads to an incorrect result.

lib/assessment.js
```diff
@@ -5,7 +5,15 @@
...
+        minimum = arguments[i];
+    };
 
+  };
+
+  return minimum;
```
srhoulam, line 16: With the above bug fixed, this is a good solution. 9/10

lib/assessment.js
```diff
@@ -13,14 +21,23 @@ var min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 var sum = function sum() {
-
+  var total = arguments[0];
```
srhoulam, line 24: The following loop will add all elements of `arguments` to `total`, including the 0th element. This causes the 0th element to be added to the sum twice. This can be corrected by initializing `total` to `0`.

lib/assessment.js
```diff
@@ -13,14 +21,23 @@ var min = function min() {
...
+  for (var i = 0; i < arguments.length; i++) {
+    total += arguments[i];
+  };
+  return total;
```
srhoulam, line 28: Correct save for a small bug. 9/10

lib/assessment.js
```diff
@@ -13,14 +21,23 @@ var min = function min() {
...
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
+var array = [];
+  for (var i = 0; i < length; i++) {
+//I am unsure how to set all the elements to the defaultValue.
```
srhoulam, line 37: Assign the default value to the array at index `i`. This way, it will assign it to a different index in each loop iteration.

lib/assessment.js
```diff
@@ -13,14 +21,23 @@ var min = function min() {
...
+  };
 
+return array;
```
srhoulam, line 40: You have most of a correct solution. 3/4

lib/assessment.js
```diff
@@ -29,7 +46,11 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+var array = [];
+  for (var i = 0; i < length; i++) {
+ // again I am unsure here how to pass the defaultFunction index of the array
```
srhoulam, line 51: You pass a function an argument by placing it inside the invocation operator (`(` `)`) when you invoke it. You're iterating from zero to `length`, so the index of the array here is `i`.

lib/assessment.js
```diff
@@ -29,7 +46,11 @@ var newArray = function newArray(length, defaultValue) {
...
+  };
+  return array;
```
srhoulam, line 53: You did most of the tricky stuff like in Q3, but sat out the part that makes this function most interesting. 1/2

lib/assessment.js
```diff
@@ -39,7 +60,12 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
+ // I am finding this question hard to break down into smaller steps to start this work. I know I need to fun an if function and I know I need a return but not sure what else.
```
srhoulam, line 63: 1. Determine whether an object was passed in. Create one if not.
2. Assign the passed-in `method` to the object's property of name `propertyName`
3. Return the object

srhoulam:
3.05/5

PR #32, opened 2015-10-18
Assessment resubmitted...

jcmeyer10:
...with Q4 only question I was asked to re answer without any sample solutions or prompts.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -5,15 +5,19 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
+  for (i=0; i< arguments.length; i++){
+    return arguments+
+  }
+min(1,4,9);
 
-};
+// I know this is wrong, I need to put args in min(args) but cant remember how to set it to the next variable because when I type in arg = arguments, arugments is already defined and my code will only accept three cases right now because of how its set up as a simple if/then statement.
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
-
+var sum = function sum(arg1,arg2) {
+  return arg1 + arg2;
 };
 
 // Question 3
@@ -21,6 +25,7 @@ var sum = function sum() {
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
 
+//didn't even get this far, just need to see it again.  couldn't find antony's code along.  can you please update any repo with the code along so if we go over it again, it's not a blank with the starter code.
 };
 
 // Question 4
@@ -32,6 +37,8 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 
 };
 
+didn't even get this far, just need to see it again.  couldn't find antony's code along.  can you please update any repo with the code along so if we go over it again, it's not a blank with the starter code.  i will use my lunch time or after the day is over to go over this.  just let me know.
+
 // Questions 5
 // Write a function that adds or replaces a method on an object
 // using the given propertyName
@@ -41,6 +48,7 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 var addMethod = function addMethod(propertyName, method, obj) {
 
 };
+/didn't even get this far, just need to see it again.  couldn't find antony's code along.  can you please update any repo with the code along so if we go over it again, it's not a blank with the starter code.
 
 // Used by testing
 module.exports = {
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -4,11 +4,15 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-var min = function min() {
-  for (i=0; i< arguments.length; i++){
-    return arguments+
+var min = function () {
+  var minimum = arguments[0]
+  for (i=1; i< arguments.length; i++){
+      if (arguments[i] < minimum) {
+        mimumun = arguments][i];
+    }
   }
-min(1,4,9);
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -4,23 +4,39 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-var min = function min() {
+var min = function () {
+  var minimum = arguments[0]
+  for (i=1; i< arguments.length; i++){
```
srhoulam, line 9: `i=1` needs a `var` to be safe! This overwrite a variable `i` in a surrounding scope, or, worst-case, create a global variable `i`!

lib/assessment.js
```diff
@@ -4,23 +4,39 @@
...
+      if (arguments[i] < minimum) {
+        mimumun = arguments][i];
```
srhoulam, line 11: Typos are some of the most insidious of bugs. They can break a lot of functionality and usually are too small to see without squinting. There is a typo on each side of the assignment operator (`=`) in this statement.

lib/assessment.js
```diff
@@ -4,23 +4,39 @@
...
+    }
+  }
+  return minimum;
+}
```
srhoulam, line 15: Great. Typos aside, this does what it needs to do. 1

lib/assessment.js
```diff
@@ -4,23 +4,39 @@
...
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
-
+var sum = function sum(arg1,arg2) {
+    var total = 0;
+    for (var i = 0; i < arguments.length; i++){
+      total += arguments[i];
+    }
+    return total;
```
srhoulam, line 28: Right on. This function takes named parameters, but actually accesses its parameters through the `arguments` pseudo-array. So it works for any number of addends. 1

lib/assessment.js
```diff
@@ -29,7 +45,11 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  array2 = [];
```
srhoulam, line 48: This is assigning to a variable `array2` found outside this function's scope or creating a global variable called `array2`.

lib/assessment.js
```diff
@@ -29,7 +45,11 @@ var newArray = function newArray(length, defaultValue) {
...
+  for (var i = 0; i < defaultsFunction.length; i++){
+    array2 = array2[defaultValue]
```
srhoulam, line 50: This is assigning the value stored in `array2[defaultValue]` into `array2`, causing us to lose the reference to the array in the very first iteration.

Also, there is no local `defaultValue`, so it's grabbing it from an outside scope. There is no accounting for what it might be.

lib/assessment.js
```diff
@@ -29,7 +45,11 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  array2 = [];
+  for (var i = 0; i < defaultsFunction.length; i++){
```
srhoulam, line 49: You are counting up here to `defaultsFunction.length` rather than the passed-in `length` parameter. A function's `length` attribute is the number parameters it expects (its arity) in case you're curious.

lib/assessment.js
```diff
@@ -29,7 +45,11 @@ var newArray = function newArray(length, defaultValue) {
...
+    array2 = array2[defaultValue]
+  }
+  return array2;
```
srhoulam, line 52: 1/4

lib/assessment.js
```diff
@@ -38,8 +58,13 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
 // The function you write should return
 // the object passed in as an argument or the object created
-var addMethod = function addMethod(propertyName, method, obj) {
 
+var addMethod = function (propertyName, method, obj) {
+  if (!obj) {
+    obj = {};
+  }
+  obj[propertyName] = method;
+  return obj;
```
srhoulam, line 67: Excellent. 1

lib/assessment.js
```diff
@@ -4,23 +4,39 @@
...
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-
+  var array = [];
+  for (var i = 0; i < arguments.length; i++){
+    arrray[i] = defaultValue;
```
srhoulam, line 37: A typo here on the left side of the assignment operator (`=`). Can you spot it?

lib/assessment.js
```diff
@@ -4,23 +4,39 @@
...
+    return total;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-
+  var array = [];
+  for (var i = 0; i < arguments.length; i++){
```
srhoulam, line 36: This loop is counting up to the wrong number. There is a `length` parameter to this function that is intended to tell it how long to make the array. However, your loop that populates the array counts up to the number of arguments passed in. Under correct usage, this function would always produce an array of length 2.

lib/assessment.js
```diff
@@ -4,23 +4,39 @@
...
+    arrray[i] = defaultValue;
+  }
+  return array;
```
srhoulam, line 39: 1/2

srhoulam:
3.75/5

PR #34, opened 2015-10-19
Lesson

julie1013:


lib/assessment.js
```diff
@@ -5,22 +5,25 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
-
+  return 3;
```
srhoulam, line 8: This always returns 3.

lib/assessment.js
```diff
@@ -5,22 +5,25 @@
...
 };
+min();
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
+var sum = function sum(a, b) {
```
srhoulam, line 16: This takes strictly two arguments and sums them. 1/2

lib/assessment.js
```diff
@@ -5,22 +5,25 @@
...
 
 };
+sum(3, 4);
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-
+  //I DON'T KNOW
```
srhoulam, line 26: This is a great exercise in breaking things in to pieces:
1. Create a new array and store it in a variable.
2. Loop from 0 to `length` and store `defaultValue` in the new array's `i`th index.
3. Return the variable storing the array.

lib/assessment.js
```diff
@@ -29,7 +32,10 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-
+  var defaultsFunction = function (){
+    ///??????? I DON'T KNOW!!!!!!!!!
```
srhoulam, line 36: This one is different from the previous question in step 2:
1. Loop from 0 to `length` and store the return value of `defaultsFunction` invoked on `i` in `array[i]`.

lib/assessment.js
```diff
@@ -39,9 +45,19 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
-
+  var Method = {
+    propertyName: propertyName,
+    method: method,
+    obj: obj
+  };
```
srhoulam, line 52: This builds a new object that has a reference to both `obj` and `method`. It doesn't satisfy the question, but it does do something important right: it places a method on some object.

Here's how I would have solved this:

``` js
  if(!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
}
```

1/2

srhoulam:
1/5

PR #38, opened 2015-10-26
Assessment

pguzman26:
Question 5 should be explained to a point that is understandable. It was so hard to interpret the following 

would of made it much easier if you said ( if the obj has not been created then create a new one.

### DIFF
--- a/lib/assessment.js
+++ b/lib/assessment.js
@@ -5,7 +5,12 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
-
+  var result = argument[0];
+    for(var i =1; i < argument.legth; i++) {
+      if(argument[i] < result) {
+        resutl = arguement[i];
+      }
+    }
 
 };
 
@@ -13,21 +18,25 @@ var min = function min() {
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
-  var one = number(one);
-  var two = number(two);
-  return number(one+two);
+var sum = function() {
+  var result = 0;
+  for (var i = 0; i < argument.length; i++) {
+    result += argument[i];
+  }
 
+return result;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
-  var length = this.length;
+  var array =[];
+    for (var i = 0; i < length; i++) {
+      array[i] = defaultValue;
+    }
+      return array;
 
-
-return newArray;
 };
 
 // Question 4
@@ -36,8 +45,11 @@ return newArray;
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
-  length = [];
-
+  var array = [];
+  for (var i = 0; i < length; i++) {
+    array[i] = defaultsFunction(i);
+  }
+return array;
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/assessment.js
```diff
@@ -5,6 +5,12 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 var min = function min() {
+  var result = argument[0];
```
srhoulam, line 8: Replace all instances of `argument` with `arguments`, and this solution works.

lib/assessment.js
```diff
@@ -5,6 +5,12 @@
...
+    for(var i =1; i < argument.legth; i++) {
```
srhoulam, line 9: There's a typo in the condition that will make this function fail in subtle and strange ways.

lib/assessment.js
```diff
@@ -5,6 +5,12 @@
...
+      if(argument[i] < result) {
+        resutl = arguement[i];
```
srhoulam, line 11: Two typos here.

lib/assessment.js
```diff
@@ -5,6 +5,12 @@
...
+      }
+    }
```
srhoulam, line 13: This function is good in most of the ways that matter most, but it needs some fixing. Most important thing to get out of this is that the way to access arguments inside a function is the `arguments` pseudo-array. It's called `arguments`, plural, because it's expected to have more than one. I'll only count that mistake once since you did this all in one session. 3/5

lib/assessment.js
```diff
@@ -12,14 +18,24 @@ var min = function min() {
...
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
+var sum = function() {
+  var result = 0;
+  for (var i = 0; i < argument.length; i++) {
+    result += argument[i];
+  }
 
+return result;
```
srhoulam, line 27: Good. 1

lib/assessment.js
```diff
@@ -12,14 +18,24 @@ var min = function min() {
...
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
+  var array =[];
+    for (var i = 0; i < length; i++) {
+      array[i] = defaultValue;
+    }
+      return array;
```
srhoulam, line 38: Great. Fix your indentation, though. 1

lib/assessment.js
```diff
@@ -29,6 +45,11 @@ var newArray = function newArray(length, defaultValue) {
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
+  var array = [];
+  for (var i = 0; i < length; i++) {
+    array[i] = defaultsFunction(i);
```
srhoulam, line 50: Correct!

lib/assessment.js
```diff
@@ -29,6 +45,11 @@ var newArray = function newArray(length, defaultValue) {
...
+  }
+return array;
```
srhoulam, line 52: 1

lib/assessment.js
```diff
@@ -38,10 +59,19 @@ var newArray2 = function newArray2(length, defaultsFunction) {
...
 // the object passed in as an argument or the object created
-var addMethod = function addMethod(propertyName, method, obj) {
+var addMethod = function(propertyName, method, obj) {
+
+if (!obj) {
+  obj = {};
+}
+obj[propertyName] = method;
+
+return obj;
```
srhoulam, line 69: Right on. 1

srhoulam:
4.6/5

PR #40, opened 2015-10-28
submitting corrected answers for JS functions ins & outs

christinemoses:


lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+// };
+
+// I am not sure how to write this correctly, but I would use Math.min to find the minimum number in an array.
+
+
+// correct answer:
+var min = function(){
+  var result = arg [0]; // first spot in the array - this is where to store the result
+  for(var index; index < arguments.length; index++); // iterate over the array
+    if(arguments[index] < result);
```
srhoulam, line 24: `for`s and `if`s must be followed by statements in order to do anything meaningful. If they're followed by semicolons, they do nothing; in the case of a `for` loop, it does nothing several times, incrementing the index variable as it goes.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+      result = arguments[index];
```
srhoulam, line 25: As a result, this line runs unconditionally and only once. It runs with the final value of `index` that caused the loop to stop, which is `arguments.length`. As a result, `result` will always be set to `undefined` when `min` is invoked.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+}
+return result; // have to see everything in the result, so return has to be outside the function
```
srhoulam, line 27: The above curly brace matches the one from line 21, so this `return` statement is taking place outside of a function. `return` only works when inside a function.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+//   };
+
+// };
+
+// I am not sure how to write this correctly, but I would use Math.min to find the minimum number in an array.
+
+
+// correct answer:
+var min = function(){
+  var result = arg [0]; // first spot in the array - this is where to store the result
```
srhoulam, line 22: `arguments`, not `arg`, is what you want. It's how you access unnamed parameters from inside a function.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+
+//   return sumTotal;
+
+// };
+
+// correct answer:
+var sum = function sum(){
+  var result = 0;
+  for (var index = 0; index < args.length; index ++);
+    result += arg[index];
```
srhoulam, line 51: Supposing this referenced `arguments`, the same problem as above would apply. The loop would do nothing `arguments.length` times, `index` would contain the value `arguments.length` when this line runs, and this line would assign `0 + undefined` (`NaN`) to `result`.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+}
+return result;
```
srhoulam, line 53: This is another `return` from outside a function. It will do nothing but throw a syntax error.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+// var newArray = function newArray(length, defaultValue) {
 
+// };
+
+//corrected answer:
+var newArray = function newArray(length,defaultValue){
+  var result = [];
+  // for loop is to do something a certain number of  times
+  for (var i = 0; i < length; i++) { // length is the size of the new array
+    result.push(defValue);
```
srhoulam, line 68: This statement is within the body of the `for` loop; it will be run multiple times. Well done. However, no variable called `defValue` exists in this function's scope.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+  };
 };
+return result;
```
srhoulam, line 71: As noted previously, this line will not work.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
 // defaultsFunction should be passed the index of the array element to be set
-var newArray2 = function newArray2(length, defaultsFunction) {
+// var newArray2 = function newArray2(length, defaultsFunction) {
+
+// };
 
+// corrected answer:
+var newArray2 = function newArray2(length, defaultsFunction){
+  for(index = 0; i < length; i++){
+    array[i] = defaultsFunction(i);
```
srhoulam, line 85: No variable called `array` exists in this function's scope. Good invocation of `defaultsFunction` to the right of the assignment operator.

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
+  };
+return array;
```
srhoulam, line 87: This `return` is within the function. It will return whatever is stored in `array`, if the variable exists.

1/2

lib/assessment.js
```diff
@@ -4,44 +4,111 @@
...
-var addMethod = function addMethod(propertyName, method, obj) {
 
+// corrected answer:
+var addMethod = function addMethod(propertyName, method, obj) {
+  if (obj === undefined){
+    obj = {}; // this assigns a value to obj if it's undefined
+  };
+  obj[propertyName] = method; // method is storing a function
+  // this takes the object passed in as an argument
+return obj; // return inside the function in this case
```
srhoulam, line 106: This is exactly correct.

`return` inside the function in **every** case. You either write your `return` statement inside a function's body, or you don't write a `return` statement at all.

1

srhoulam:
1.5/5

PR #42, opened 2015-10-29
Lesson

julie1013:


lib/assessment_redux_10_28.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+
+// Question 1
+// Write a function that takes zero or more arguments
+// and returns the minimum from that set
+// assume that all the argument are numbers
+var min = function() {
+  var result = arguments[0];
+  for (i = 1; i < arguments.length; i++){
```
srhoulam, line 9: `i = 1` here is an assignment to an undeclared variable. The interpreter will look for `i` in outer scopes, until it reaches the global scope, and if it doesn't find it there, it will create it. If it does find it somewhere, it will **overwrite it** with 1, interfering with code elsewhere in your program. This is solved by replacing the assignment statement `i = 1` with the variable declaration and initialization `var i = 1`.

lib/assessment_redux_10_28.js
```diff
@@ -0,0 +1,83 @@
...
+    if (arguments[i] < result){
+      result = arguments[i];
+    }
+
+  }
+return result;
```
srhoulam, line 15: Good `min` function. 9/10

lib/assessment_redux_10_28.js
```diff
@@ -0,0 +1,83 @@
...
+// Question 2
+// Write a function that takes zero or more arguments
+// and returns the sum of that set
+// assume that all the argument are numbers
+var sum = function() {
+  var result = 0;
+  for (i = 0; i < arguments.length; i++){
+    result = arguments[i] + result;
+  }
+return result;
```
srhoulam, line 29: Good `sum` function. Same trouble with `i` as above. 9/10

lib/assessment_redux_10_28.js
```diff
@@ -0,0 +1,83 @@
...
+
+// Question 3
+// Write a function that returns an array of the specified length
+// with all the elements set to defaultValue
+var newArray = function(length, defVal){
+  var result = [];
+  for (i = 0; i < length; i++){
+    result.push(defVal);
+  }
+return result;
```
srhoulam, line 41: Great. 9/10

lib/assessment_redux_10_28.js
```diff
@@ -0,0 +1,83 @@
...
+
+// Question 4
+// Write a function that returns an array of the specified length
+// with all the elements set to the value returned from
+// invoking the argument defaultsFunction
+// defaultsFunction should be passed the index of the array element to be set
+var newArray2 = function(length, defFunction){
+  var result = [];
+  for (i = 0; i < length; i++){
+    result.push(defFunction);
```
srhoulam, line 53: This is populating an array with references to `defFunction`; a user can then call `defFunction` from the array indices it is now assigned to.

This question wanted you to invoke the function on the current index and populate the array with its return values. That means you need to invoke `defFunction` in the value expression that is being added to the array. A user would not be able to invoke `defFunction` from the values in the array unless `defFunction` returned itself for some index.

lib/assessment_redux_10_28.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+return result;
+
+  };
```
srhoulam, line 58: Almost. 3/5

lib/assessment_redux_10_28.js
```diff
@@ -0,0 +1,83 @@
...
+// If the argument obj is unset, create a new object
+// The function you write should return
+// the object passed in as an argument or the object created
+var addMethod = function(propertyName, method, obj){
+  if (obj === undefined){
+    obj = {};
+  }
+  obj[propertyName] = method;
+  return obj;
+  };
```
srhoulam, line 73: Great. 1

srhoulam:
4.3/5

PR #44, opened 2015-10-29
Response

leahphuong:


lib/assessment.js
```diff
@@ -4,31 +4,62 @@
...
+var min = function min(nums) {
+  //grab the first number to compare against
+  var minValue = nums[0];
+  // loop through the remaining numbers
+  for(var i = 1; i < nums.length; i++){
+    if(nums[i] < minValue){
+      //replace minValue with new number
+      minValue = nums[i];
+    }
+    return minValue;
```
srhoulam, line 16: Returning unconditionally from within a loop means it's not a loop anymore. These statements run exactly once. This function actually finds the minimum of the first two elements of an array.

lib/assessment.js
```diff
@@ -4,31 +4,62 @@
...
+  }
```
srhoulam, line 17: If you replace all instances of `nums` with `arguments` and move the `return` statement out of the loop body, this will be a correct solution. 1/2

lib/assessment.js
```diff
@@ -4,31 +4,62 @@
...
 // and returns the sum of that set
 // assume that all the argument are numbers
-var sum = function sum() {
 
+var sum = function sum(nums) {
+  var sumValue = 0;
+  for(var i = 0; i < nums.length; i++){
+      sumValue += nums[i];
+  }
+return sumValue;
```
srhoulam, line 31: Great. Like above, replace all occurrences of `nums` with `arguments`, and it's solid. 9/10

lib/assessment.js
```diff
@@ -4,31 +4,62 @@
...
 };
 
+
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 var newArray = function newArray(length, defaultValue) {
+  //create a new empty array
+  //loop through length and push default Value to the new empty arrays
+    var Array = [];
```
srhoulam, line 41: `Array` is a built-in constructor function the interpreter provides. It will not allow you to create a variable with that name. It will throw an error with a very strange message.

lib/assessment.js
```diff
@@ -4,31 +4,62 @@
...
+    for(var i = 0; i < length; i++){
 
+     Array[i] = defaultValue;
+   }
+    return Array;
```
srhoulam, line 46: If you had named your new array `array` instead of `Array`, this function would work correctly. 9/10

lib/assessment.js
```diff
@@ -4,31 +4,62 @@
...
 // invoking the argument defaultsFunction
 // defaultsFunction should be passed the index of the array element to be set
 var newArray2 = function newArray2(length, defaultsFunction) {
+// inside the for loop, invoke defaultfunction and push the
+// default value to the new array
+    var Array = [];
+    for(var i = 0; i < length; i++){
+      Array[i] = defaultsFunction(i);
+    }
+  return Array;
```
srhoulam, line 62: Same as above. 9/10

lib/assessment.js
```diff
@@ -39,7 +70,11 @@ var newArray2 = function newArray2(length, defaultsFunction) {
 // The function you write should return
 // the object passed in as an argument or the object created
 var addMethod = function addMethod(propertyName, method, obj) {
-
+  if(obj === undefined){
+    var obj = {};
+    }
+    obj[propertyName] = method;
+    return obj;
```
srhoulam, line 77: Yep. 1

srhoulam:
4.2/5

PR #51, opened 2016-01-26
JS In & Out Diagnostic

msyao:


lib/diagnostic.js
```diff
@@ -5,23 +5,34 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let args = Array.isArray(arguments[0]) ? arguments[0] : arguments;
```
GA-MEB, line 8: This is not necessary - "assume that all the argument are numbers"

lib/diagnostic.js
```diff
@@ -5,23 +5,34 @@
...
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
-};
+const sum = function sum(array) {
+  let sum = array[0];
```
GA-MEB, line 23: In this example, you should just use `arguments`.

lib/diagnostic.js
```diff
@@ -5,23 +5,34 @@
...
+  for (let i =0; i <array.length, i++){
+    sum += array[i++];
```
GA-MEB, line 25: The additional `i++` will cause this to not work, because you're already incrementing `i` as part of the `for` loop.

GA-MEB:
1.5 / 5 (-0.5 on Q2)

Let me know if you'd like to go over this or resubmit it - happy to give you more feedback.

PR #52, opened 2016-01-26
The Diagnostic

MFBrewster:


lib/diagnostic.js
```diff
@@ -13,14 +17,21 @@ const min = function min() {
...
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let ary = new Array(length);
+  return ary.map(function(element, index)  {
+    ary[index] = typeof defaultValue === 'function' ? defaultvalue(index) : defaultvalue;
```
GA-MEB, line 33: Interesting - I don't think something called `defaultValue` is likely to be a function, but this is a reasonable way to handle that case.

[EDIT] Sorry, I didn't realize this was a combined answer for Q3 and Q4.

lib/diagnostic.js
```diff
@@ -36,7 +48,9 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
-
+  let returnedObj = obj || new Object();
+  returnedObj.propertyName = method;
```
GA-MEB, line 52: Good thought, but not quite right. This will literally set a property called `propertyName`, not whatever String you would have passed in as an argument.

lib/diagnostic.js
GA-MEB, line 33: Unfortunately, you didn't quite use `map` correctly. When `map` takes a callback function, the _return value_ of that callback gets automatically assigned as the corresponding value in the new array. As written, `map` will return an array of `undefined` values because nothing is returned.

GA-MEB:
3.5 / 5

Let me know if you have any questions, or want further feedback.

PR #54, opened 2016-01-26
Finish diagnostic

dumplingpuff:


lib/diagnostic.js
```diff
@@ -20,7 +31,18 @@ const sum = function sum() {
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let newArr = [];
+  if(defaultValue instanceof Array) {
```
GA-MEB, line 35: I don't know that `defaultValue` is necessarily only an array or a function - it could easily be a number or a string.

lib/diagnostic.js
```diff
@@ -20,7 +31,18 @@ const sum = function sum() {
...
+    for(let i = 0; i < length; i++) {
+      newArr[i] = defaultValue[i];
+    }
+    return newArr;
```
GA-MEB, line 39: This doesn't really address what `newArray` is supposed to do - create new array of size `length`, populated with some default value.

lib/diagnostic.js
```diff
@@ -20,7 +31,18 @@ const sum = function sum() {
...
+  }
+  else if (defaultValue instanceof Function) {
+    array.forEach(function(element, index, array) {
+    array[index] = callback(element,index, array);
```
GA-MEB, line 43: Where is `callback` defined? Did you mean to call the `defaultValue` function?

GA-MEB, line 43: Why not just pass in the `defaultValue` function as the callback for `.forEach`?

lib/diagnostic.js
```diff
@@ -36,7 +58,11 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
-
+  if(typeof obj[propertyName] === 'undefined') {
+    obj[propertyName] = method();
+    return true;
+  }
+  return false;
```
GA-MEB, line 65: This is not the correct behavior. If `obj` is undefined, the function should return a new object with a single property; that property's name is the value of `propertyName`, and its value is the method that was passed in.

GA-MEB, line 65: If `obj` is defined, that key-value pair should be added or overwritten.

GA-MEB:
2.75 / 5 (-0.75 on Q3, Q4, Q5)

Let me know if you have any questions or want additional feedback.

PR #55, opened 2016-01-26
A.Jordon diag Response. Couldnt complete question 5.

ajordon:


lib/diagnostic.js
```diff
@@ -28,6 +46,21 @@ const newArray = function newArray(length, defaultValue) {
 // function then the elements in the array will be set
 // to the result of invoking defaultValue with the
 // elements index as an argument
+const newArray = function newArray(length, defaultValue) {
+  let ary = new Array(length);
+  if (typeof defaultValue === 'function') {
+    for (let i =0; i < length; i++) {
+      ary.forEach(defaultValue[i]);
```
GA-MEB, line 53: You're not using `forEach` correctly. It expects a function as an argument, and executes that function once for each element in the array; additionally, it takes that element as an argument. If you want to use `forEach` in this problem, it would be _instead_ of your `for` loop.

Since you're already using a `for` loop here, all you need to do is write `ary[i] = defaultValue(i)`, much like you wrote `ary[i] = defaultValue` below.

lib/diagnostic.js
```diff
@@ -28,6 +46,21 @@ const newArray = function newArray(length, defaultValue) {
...
+    }
+  }
+  else {
+    for (let i =0; i < length; i++) {
+      ary[i] = defaultValue);
```
GA-MEB, line 58: Watch out for extra characters!

lib/diagnostic.js
```diff
@@ -36,7 +69,17 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
+  let newObj = {};
+  newObj = obj;
```
GA-MEB, line 73: Not quite. If no object is passed in through `obj`, this will set `newObj` to `undefined`.

lib/diagnostic.js
```diff
@@ -36,7 +69,17 @@ const newArray = function newArray(length, defaultValue) {
...
+
+  if (obj.propertyName !== null) {
+    newOnj.propertyName = obj.propertyName;
```
GA-MEB, line 76: This will try to change a property called 'propertyName' rather than the property name that gets passed in as an argument.

GA-MEB:
3.75 / 5 (-0.5 on Q4, -0.75 on Q5)

Let me know if you have any questions, want further feedback, or are planning to resubmit!

PR #56, opened 2016-01-26
Functions, Ins & Outs

kenvng:
I need help, Please!!!

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,22 +5,33 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let answer = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    if (answer < arguments[i]) {
+      answer = arguments[i];
+    }
+  }
+  return(answer);
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(x, y) {
+  return x + y;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let newArray = function(length, defaultValue) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,22 +5,33 @@
...
 // assume that all the argument are numbers
 const min = function min() {
-
+  let answer = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    if (answer < arguments[i]) {
+      answer = arguments[i];
+    }
+  }
+  return(answer);
```
GA-MEB, line 14: As written, this function will return the **largest** value in the set, not the smallest. The trick is that if any given element is smaller than the _smallest so far_ (which is really what we're storing inside `answer`) it becomes the new smallest element.

lib/diagnostic.js
```diff
@@ -5,22 +5,33 @@
...
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(x, y) {
+  return x + y;
```
GA-MEB, line 22: This function does not handle a variable number of arguments.

lib/diagnostic.js
```diff
@@ -5,22 +5,33 @@
...
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let newArray = function(length, defaultValue) {
+    for (let i = 0; 1 < newArray.length; i++) {
+      newArray[i] = defaultValue;
+    }
+  };
+return newArray[i];
```
GA-MEB, line 34: Partially right - if you had actually made `newArray` a new array, and returned `newArray`, this would have worked.

GA-MEB:
1 / 5 (-0.5 on Q1, -0.72 on Q2, -0.75 on Q3)

Let me know if you'd like to walk through the answers together - we can meet up for a couple of minutes at the end of the day.

kenvng:
Please, I really need help.  I would love the opportunity to meet up with
you and you can go through them with me.  Thank you.

PR #57, opened 2016-01-26
Answers to diagnostic

kpie89:


lib/diagnostic.js
```diff
@@ -5,7 +5,15 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+    let args = Array.isArray(arguments[0]) ? arguments[0] : arguments;
```
GA-MEB, line 8: This is not necessary - "assume that all the argument are numbers"

lib/diagnostic.js
```diff
@@ -13,22 +21,35 @@ const min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
+    let args = arguments[0] instanceof Array ? arguments[0] : arguments;
```
GA-MEB, line 24: "assume that all the argument are numbers"

lib/diagnostic.js
```diff
@@ -13,22 +21,35 @@ const min = function min() {
...
 
+  return answer;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  newArray.forEach(length, defaultValue);
```
GA-MEB, line 37: No, this isn't the correct use of `.forEach`.

lib/diagnostic.js
```diff
@@ -13,22 +21,35 @@ const min = function min() {
...
 };
 
+return newArray;
```
GA-MEB, line 40: Why is your return value outside of the function?

lib/diagnostic.js
```diff
@@ -13,22 +21,35 @@ const min = function min() {
...
 // Add code to newArray so that if defaultValue is a
 // function then the elements in the array will be set
 // to the result of invoking defaultValue with the
 // elements index as an argument
+const newArray = function newArray(length, defaultValue) {
+  if (defaultValue())
+    newArray.forEach(length, defaultValue);
+};
 
+return newArray;
```
GA-MEB, line 52: See notes above.

lib/diagnostic.js
```diff
@@ -36,7 +57,10 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
-
+  if (typeof obj [method] === 'undefined') {
+  obj[method] = propertyName;
+  return true;
```
GA-MEB, line 62: This function does not return the modified object. Additionally, you've confused `propertyName` (the key) and `method` (the value). Finally, you haven't accounted for what happens if `obj` is undefined (i.e. no value is passed in as a third argument).

GA-MEB:
2 / 5 (-1 on Q3, Q4, Q5)

Let me know if you have any questions, would like more feedback, or want to resubmit.

PR #60, opened 2016-01-26
struggled with this

nwilcox11:


RealWeeks:
2.5/5.  I can see you were a bit challenged here.  Looks like it was a time limit issue, is that correct? What could you use more clarification on. Please response do I have a sense of your understanding, so we know what to cover or review.

PR #62, opened 2016-01-26
Response to DIA

edesilets:


RealWeeks:
2.5/5.  Great work with min and sum.  It looks like this was more of a time issue.  Do some code wars to level up your skill with functions.

Also looks like you tried to open up a block with .push.  If there is ever a question with a method feel free to look up the documentation.

PR #63, opened 2016-01-26
wrote some of the answer

Knguyen21:


RealWeeks:
3.5/5.  Looks like you ran out of time.  You didn't set the default values.  If you run grunt test you can see where the tests failed. This seems to be more of an issue of time than skill.  I would recommend some code wars to level up your functions skills.

PR #64, opened 2016-01-26
Complete 2 questions

seatuna:


RealWeeks:
1/5. Your min works, but your sum doesn't. You used "args" instead of arguments.  Also run it in node and console log your result.  Does it return the predicted answer.  You need a bit more practice with functions.

PR #65, opened 2016-01-26
Response to diagnostic

kylekatz:


RealWeeks:
1/5. Kyle I think you need some function review.  Also run your code in node before you submit and check for blowups.  Don't get discouraged, I think you need to so some work on functions fundamentals.  Maybe some node school functional javascript or closures.

Your if else statements are incorrect causing a blowup.  I'm also confused why you did Math.min

In question 3 your want to return the result not attempt to modify the stored function, causes a blow up.

Quest 4 new array2 isn't defined.

PR #66, opened 2016-01-26
answers

jdamiron:


RealWeeks:
2/5.  Min and sum are correct but not the other functions.  Compare your answer against the solution branch and let me know if you have any questions.  Do some code wars to level up your skills on this.

PR #69, opened 2016-03-28
Reply to diagnostic js-func-in-and-outs

cuprous:


lib/diagnostic.js
```diff
@@ -5,7 +5,20 @@
...
+    minimum = arguments[0];
+  }
 
+  for (let i = 1; i < arguments.length; i++) {
+    if (arguments[i] < minimum) {
+      minimum = arguments[i];
+    }
+  }
+
+  return minimum;
```
faetea, line 21: looks good

lib/diagnostic.js
```diff
@@ -13,14 +26,36 @@ const min = function min() {
...
+    numHolder = undefined;
+  } else {
+    numHolder = arguments[0];
+  }
+
+  for (let i = 1; i < arguments.length; i++) {
+    numHolder += arguments[i];
+  }
 
+  return numHolder;
```
faetea, line 40: looks good.  it isn't necessary to have the if/else on either question.  you can just write `let numHolder = arguments[0];`

lib/diagnostic.js
```diff
@@ -13,14 +26,36 @@ const min = function min() {
...
+    for (let i = 0; i < length; i++) {
+      arrayCopy.push(defaultValue(i));
+    }
+  } else {
+    for (let i = 0; i < length; i++) {
+      arrayCopy.push(defaultValue);
+    }
+  }
 
+  return arrayCopy;
```
faetea, line 58: so instead of having two for loops inside of the if/else it would make more sense to have one for loop around the if/else.  like this:

```
let arrayCopy = [];
    if (typeof defaultValue === 'function') {
        arrayCopy.push(defaultValue(i));
    } else {
        arrayCopy.push(defaultValue);
    }
}
return arrayCopy;
```

lib/diagnostic.js
```diff
@@ -36,7 +71,17 @@ const newArray = function newArray(length, defaultValue) {
...
+  if (typeof obj === 'object') {
+    obj[propertyName] = method;
+    return obj;
+
+  } else {
+    newObj = {};
+    newObj[propertyName] = method;
+    return newObj;
 
+  }
```
faetea, line 84: looks good.  you could also write it as:

```
if (!obj) {
    obj = {};
}
obj[propertyName] = method;
return obj;
```

faetea:
4/4

Great effort!

PR #70, opened 2016-03-28
Complete js-functions-in-and-out

philschoof:


lib/diagnostic.js
```diff
@@ -5,7 +5,13 @@
...
 // assume that all the argument are numbers
 const min = function min() {
-
+    let result = arguments[0];
+    for (let i = 1; i < arguments.length; i++) {
+      if(arguments[i] < result){
+        result = arguments[i];
+      }
+    }
+  return result;
```
faetea, line 14: looks good

lib/diagnostic.js
```diff
@@ -13,14 +19,33 @@ const min = function min() {
...
+  let result = 0;
+  if (arguments.length > 0){
+    for (let i = 0; i < arguments.length; i++) {
+      result += arguments[i];
+    }
+  return result;
+}else{
+  let result = 0;
+  return result;
+}
```
faetea, line 31: We don't need an if/else in this function, it will create an error.  Also we want to start with `let result = arguments[0];`

lib/diagnostic.js
```diff
@@ -13,14 +19,33 @@ const min = function min() {
...
+  if (typeof defaultValue === "function"){
+    for (let i = 0; i < length; i++){
+      result[i] = defaultValue();
+    }
+  }else{
+    for(let i = 0; i < length; i++){
+      result[i] = defaultValue;
+    }
+  }
+  return result;
```
faetea, line 48: If we want to set result to an array ( [ ] ), then we would need to use `result.push(defaultValue())` to add to that array.  If we are going to use `result[i] = defaultValue();` in the if/else then we need to write `let result = new Array(length);`  Also, we don't need to have two for loops, we can put the if/else inside of one for loop.  Something like this:

```
  array[i];
  if (typeof defaultValue === 'function') {
       array[i] = defaultValue(i);
   } else {
        array[i] = defaultValue;
    }
}
return array;
```

lib/diagnostic.js
```diff
@@ -36,7 +63,12 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
-
+if(obj === "undefined"){
+  obj = {};
+}else{
+  obj.propertyName.method = method;
+}
+return obj;
```
faetea, line 71: This is getting declaration exceptions, because they are undefined.  Instead of checking if obj is undefined we can write `if (!obj)`.  Also, use `obj[propertyName] = method;` instead of `obj.propertyName.method = method;`, and you don't need to place it in an `else { }`.

faetea:
3 / 5

---

**Breakdown**
-.50 on Question 2
-.50 on Question 4
-1 on Question 5

PR #72, opened 2016-03-28
Response to Diagnostic

rdelvallej32:
Got stuck on the create array question. Also had trouble returning the min of all arguments.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,17 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let result = arguments[0];
 
+  for(let i = 1; i < arguments.length; i++) {
+
+    if(result < arguments[i]) {
+      result = arguments[i];
+    }
+
+  }
+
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -20,7 +37,16 @@ const sum = function sum() {
...
+  let array1 = array1[];
+
+    return function(defaultValue) {
 
+      for(let i = 0; i < length.length; i++ ) {
+        array1 = length[i];
+
+  }
+
+};
```
RDegnen, line 49:
```
  }

  return array;
};
```

lib/diagnostic.js
```diff
@@ -5,7 +5,17 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let result = arguments[0];
 
+  for(let i = 1; i < arguments.length; i++) {
+
+    if(result < arguments[i]) {
```
RDegnen, line 12: swap this line around `(arguments[i] < result)`

PR #73, opened 2016-03-28
JS In's and Outs diagnostic

pswope:
Added answers

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,23 +5,37 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i = 1; i < arguments.length; i++)
+  if (arguments[i] < 0) {
+  result = arguments[i]
+}
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,23 +5,37 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i = 1; i < arguments.length; i++)
+  if (arguments[i] < 0) {
+  result = arguments[i]
+}
```
RDegnen, line 12: all you did was forget to return result here

lib/diagnostic.js
```diff
@@ -36,8 +50,11 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
-
-};
+    if (obj === undefined) {
+      obj = {};
+    }
+    return obj;
+  };
```
RDegnen, line 57:
```
  if (!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
};
```

RDegnen:
2 out of 5 very close on one of them. Nice improvement though!

PR #75, opened 2016-03-28
js-functions-ins-and-outs-diagnostic

BenGitsCode:
Started going through codeschool heavily over the weekend and those examples
seem to be helping build my understanding a lot, but I'm still catching up much
slower than the pace of the course.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -4,24 +4,40 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
-
+const min = function min(a, b) {
+  if (a < b) {
+    return a;
+  } else if (b < a) {
+    return b;
+  }
 };
+
+// I'm not sure on how to accept zero arguments, unless its as simple as
+// leaving the () empty
+
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(a, b) {
+  return a + b;
 };
+
+
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
+  return newArray (length(defaultValue));
+
 
 };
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -4,25 +4,41 @@
...
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
-
+const min = function min(a, b) {
+  if (a < b) {
+    return a;
+  } else if (b < a) {
+    return b;
+  }
```
RDegnen, line 12:
```
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }

  return result;
};
```

lib/diagnostic.js
```diff
@@ -4,25 +4,41 @@
...
+
+
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(a, b) {
+  return a + b;
```
RDegnen, line 24:
```

  let result = arguments[0];
    result += arguments[i];
  }

  return result;
};
```

lib/diagnostic.js
```diff
@@ -4,25 +4,41 @@
...
 };
 
+
+
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
+  return newArray (length(defaultValue));
+
```
RDegnen, line 34:
```
  }

  return array;
};
```

lib/diagnostic.js
```diff
@@ -36,6 +52,7 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
+  return this.
```
RDegnen, line 55:
```
  if (!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
};
```

RDegnen:
If you need more review let me know

PR #76, opened 2016-03-28
Add info

Jcornmanhomonoff:


lib/diagnostic.js
```diff
@@ -4,23 +4,39 @@
...
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  if (Array.isArray(lenth)) {
+    let prod = 1;
+    for (let i = 0; i < length.length; i++) {
+      prod = numbers[i];
+    }
+    return prod;
```
RDegnen, line 39: watch for syntax errors here. you didn't close the if statement and wrote lenth in Array.isArray isntead of length

lib/diagnostic.js
```diff
@@ -4,23 +4,39 @@
...
-const min = function min() {
-
+const min = function min(set) {
+  let smallest = 0;
+  for (let i = 0; i < set.length; i++) {
+    if (set[i]<smallest) {
+      smallest = number[i]
+    }
+    console.log(smallest)
+  }
```
RDegnen, line 14:
```
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }

  return result;
};
```

PR #77, opened 2016-03-28
adds answers

V4Vince:


lib/diagnostic.js
```diff
@@ -36,7 +63,11 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
-
+  return {
+    propertyName;
+    method;
+    obj;
+  };
```
RDegnen, line 70:
```
  if (!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
};
```

lib/diagnostic.js
```diff
@@ -28,6 +43,18 @@ const newArray = function newArray(length, defaultValue) {
...
+  let count = i;
+  for (var i = 0; i < length; i++) {
+    if (!defaultValue) {
+      array.push(count);
+    } else {
+      array.push(defaultValue);
+    }
+  }//for
+  return array;
+};
```
RDegnen, line 57:
```
  }

  return array;
};
```

lib/diagnostic.js
```diff
@@ -5,7 +5,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i = 0; i < arugments.length; i++) {
+    if (result < arugments[i]) {
```
RDegnen, line 10: swap this around to be `(arguments[i] < result)`

RDegnen:
1 of 5

PR #78, opened 2016-03-28
JS Functions Ins & Outs Diagnostic

louarnos:
Answered Questions

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let minimum = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    if(arguments[i]< minimum){
+      minimum = arguments[i];
+    }
+  }
+  return minimum;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -36,7 +57,14 @@ const newArray = function newArray(length, defaultValue) {
...
 const addMethod = function addMethod(propertyName, method, obj) {
-
+  if(typeof obj === "object"){
+    obj.prototype.propertyName = method;
+  }
+  if(typeof arguments[2] === "undefined"){
+    obj = {};
+    obj.prototype.propertyName = method;
+  }
+  return obj;
```
RDegnen, line 67:
```
  if (!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
};
```

lib/diagnostic.js
```diff
@@ -13,14 +19,29 @@ const min = function min() {
...
+    for (let i = 0; i < length; i++) {
+      defaultArray.push(defaultValue);
+    }
+  }
+  else if((typeof defaultValue) === "object"){
+    for (let i = 0; i < length; i++) {
+      defaultArray[i] = defaultValue(defaultArray[i]);
+    }
+  }
+  return defaultArray;
```
RDegnen, line 44:
```
  }

  return array;
};
```

RDegnen:
3 out of 5 not too shabby

PR #79, opened 2016-03-28
Response to ins-outs diagnostic.

natdjerf:


lib/diagnostic.js
```diff
@@ -36,6 +64,12 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
+  let result = {};
+  for (var i = 0; i < argument.length; i++) {
+    if (argument[0]){
+      result.push(method(argument[i]));
+    };
+  }
```
RDegnen, line 72:
```
  if (!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
};
```

lib/diagnostic.js
```diff
@@ -4,31 +4,59 @@
...
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let result = [];
+  for (var i = 0; i < newArray.length; i++) {
+    result.push(defaultValue);
+  }
+  return result;
```
RDegnen, line 37: the new array should be created in the function `let array = new Array(length);`

RDegnen:
2 out of 5

PR #80, opened 2016-03-28
Complete diagnostic

olivia-leach:


lib/diagnostic.js
```diff
@@ -37,6 +57,13 @@ const newArray = function newArray(length, defaultValue) {
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
 
+  if (arguments.length < 3) {
+    let obj = {};
+  }
+
+  obj.propertyName = method;
+  return obj;
+
```
RDegnen, line 66: Sooooo close on this one. Instead of doing `obj.propertyName = method;` it should have been `obj[propertyName] = method` because propertyName is the key name you are trying to replace

RDegnen:
4 out of 5. Good job your code is looking great, so close on the last one

PR #81, opened 2016-03-28
Diagnostic, Mar. 28

susandaiute:
Answer functions ins & outs questions.  Unfinished.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -4,15 +4,25 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
-
+const min = function min(array) {
+let array = [];
+let x = array[i];
+for (x=0, x<array.length; x++) {
+	return Math.min(array);
+}
 };
+
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
+const sum = function sum(array) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -4,15 +4,25 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
-
+const min = function min(array) {
+let array = [];
```
RDegnen, line 8: your already passing in `array` as a parameter

lib/diagnostic.js
```diff
@@ -4,15 +4,25 @@
...
+let x = array[i];
```
RDegnen, line 9: theres no need to define x as `array[i]` first, i isn't defined here. And `array[i]` should only be used inside the loop where i would be defined, or in your case `x`

lib/diagnostic.js
```diff
@@ -4,15 +4,25 @@
...
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
+const sum = function sum(array) {
+  let array = [];
+  let x = array[i];
+  for (x=0;x<array.length;x++){
+    return x+sum;
+  }
```
RDegnen, line 25: same as above

RDegnen:
Do you need a review of for loops?

susandaiute:
Yup!

PR #83, opened 2016-03-28
Response to questions

ghost:


lib/diagnostic.js
```diff
@@ -4,8 +4,11 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
-
+const min = function min(sum, sum2) {
+  return Math.min.apply(sum, sum2);
+    }
```
RDegnen: This extra `}` breaks everything

lib/diagnostic.js
```diff
@@ -4,8 +4,11 @@
...
+  }
+  return min();
```
RDegnen: This also breaks everything, you are calling the function your in with no parameters and trying to return it here

RDegnen, line 11: you are return 2 different things here, and you have two `}` that are going to throw syntax errors and break everything

lib/diagnostic.js
```diff
@@ -13,14 +16,19 @@ const min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
-
+  return Math.max.apply(sum, sum2);
+    }
+  }
+  return max;
```
RDegnen, line 22: your doing the same thing as question 1 here

RDegnen:
There were syntax errors all over the place, be careful that you are properly closing your functions and loops

PR #87, opened 2016-05-23
first 2 are working

MishaHerscu:


lib/diagnostic.js
```diff
@@ -5,38 +5,86 @@
...
 const addMethod = function addMethod(propertyName, method, obj) {
-
+  if(typeof obj !== undefined){
+    obj.prototype.propertyName = method;
+  }else{
+    let obj = {};
+    obj.prototype.propertyName = method;
+  }
+  return obj;
 };
```
berziiii, line 88: When the test runs for this example, it says that there is a declaration exception issue. It also says that is cannot find prototype of property 'undefined'. This make me think it's something with a variable.

berziiii:
Nice work, 4 / 5!

PR #88, opened 2016-05-23
Answer for js-functions-ins-and-outs-diagnostic

tina513:
Answer for js-functions-ins-and-outs-diagnostic from tina

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+let mini = 0;
+for (let i = 0; i < arguments.length; i++) {
+  if(mini > arguments[i]) {
+    mini = arguments[i];
+  }
+}
+return mini;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,29 +5,54 @@
...
+  }
+  return newArr;
 };
 
 // Question 4
 // Add code to newArray so that if defaultValue is a
 // function then the elements in the array will be set
 // to the result of invoking defaultValue with the
 // element's index as an argument
+const newArray = function newArray(length, defaultValue) {
```
berziiii, line 47: You cannot reassign `newArray` since it is already defined above.

berziiii:
Since you had `const newArray` defined twice, the tests wouldn't run. When I commented out question 4, it ran and you got 3 / 5 right. Nice job. Just be careful moving forward and be sure to use `grunt nag` for syntax issues, `grunt test` to run the tests and `grunt reformat` as the README.md instructs

lib/diagnostic.js
tina513, line 47: Hey Brian,

Thank you for your grading. I changed my function name, but it still won't work. I got "

Error: Expected [ Function, Function, Function ] to equal [ 3, 2, 1 ]." error when I run "grunt test". Could you let me know what is wrong with my code? Thank you.

const newArrayNew = function newArrayNew(length, defaultValue) {
  let newArr = [];
  let i = 0;
  while (i < length) {
    newArr[i] = defaultValue(i); //this is the problem, but I don't know why
    i++;
  }
  return newArr;
};

PR #89, opened 2016-05-23
Response for ins and outs

jacobhereford:
answered diagnostic questions for ins and outs


berziiii:
Due to unexpected syntax errors, the tests wouldn't run. I advise you to revisit this diagnostic and follow the README.md where is says to us `grunt nag`, `grunt test`, and `grunt reformat`. These will help clean up the code along with point you in the right direction to solving these functions. Due to the tests not running, we cannot score at this time.

PR #91, opened 2016-05-23
JS functions ins & outs diagnostic

lcurran:


berziiii:
1 / 5 correct. Your `newArray` passes the test, but for the wrong question. What you have for question 3 actually solves number 4. But sure to run `grunt test` in the future along with `grunt nag` (like the README.md states) and it will help guide you towards what you need to think about to alter your responses.

PR #92, opened 2016-05-23
Functions ins n outs

computingbets:
Submit without grunt test. Grunt unable to work. Brian inspecting.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -4,7 +4,12 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
+const min = function min(x, y) {
+  if (x < y) {
+    return x;
+  } else {
+    return y;
+  }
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -4,39 +4,53 @@
...
+length[defaultValue] = length;
 };
 
 // Question 4
 // Add code to newArray so that if defaultValue is a
 // function then the elements in the array will be set
 // to the result of invoking defaultValue with the
 // element's index as an argument
-
+const newArray = function newArray(length, defaultValue) {
```
berziiii, line 38: Already declared this variable. you cannot redeclare.

berziiii:
Cannot score the diagnostic due to `grunt nag` issues. I advise you to revisit the diagnostic and run `grunt nag` from this you may find syntax errors that prevent `grunt test` from running. Also, when `grunt test` runs, it appears you're trying to redeclare a variable that already exists.

PR #93, opened 2016-05-23
tara favazza js-functions-ins-and-outs-diagnostic

tfavazza:


lib/diagnostic.js
```diff
@@ -5,22 +5,43 @@
...
+  return addedNumbers;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let anArray = [];
+  if(typeof defaultValue == "function"){
```
berziiii, line 34: Supposed to be `===` not `==`

lib/diagnostic.js
```diff
@@ -36,7 +57,15 @@ const newArray = function newArray(length, defaultValue) {
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod(propertyName, method, obj) {
-
+if(obj[propertyName]) {
+    obj[propertyName] = method;
+    return obj;
+  } else {
+    obj {
```
berziiii, line 64: There is a semicolon missing here. Due to this, `grunt test` won't run. Please use `grunt nag`, like the instructions state, to correct and semantic errors. This will allow `grunt test` to run correctly and identify what is passing and what is not.

berziiii:
Due to `grunt test` not running from syntax errors, we cannot score this diagnostic currently. I advise you to revisit it, clean up the syntax issues and then see if the tests run.

PR #94, opened 2016-05-23
js ins and outs diagnostic

shireen-bean:
all questions answered

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i=1;i<arguments.length;i++){
+    if (result > arguments[i]){
+      result = arguments[i];
+    }
+  }
+  return result
 };
 
 // Question 2
@@ -13,14 +19,22 @@ const min = function min() {
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
-
+  let result = 0;
+  for (let i=0;i<arguments.length;i++){
+    result += arguments[i]
+  }
+return result
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let array = [];
+  for(let i=0;i<=length-1;i++){
+    array[i] = defaultValue(i);
+  }
+return array
 };
 
 // Question 4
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,22 +5,36 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i=1;i<arguments.length;i++){
```
payne-chris-r, line 9: Make sure you start your 'for loop' from 0, not 1. There will be times that you may want to start your for loops from a number other than 0, but you almost always want to start there and build towards array.length (especially in these first few weeks)! You'll end up tripping up people working with your code--mostly `future you`!

lib/diagnostic.js
```diff
@@ -5,22 +5,36 @@
...
+    if (result > arguments[i]){
+      result = arguments[i];
+    }
+  }
+  return result
 };
```
payne-chris-r, line 15: Pay attention to your linter and/or run `grunt nag` from the console to catch typos/omitted semi colons! In addition, this function is very close to what we were looking for, but fails the `grunt test` you should be running from your console.

lib/diagnostic.js
```diff
@@ -5,22 +5,36 @@
...
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
-
+  let result = 0;
+  for (let i=0;i<arguments.length;i++){
+    result += arguments[i]
+  }
+return result
 };
```
payne-chris-r, line 28: Linter errors. Doesn't return 'undefined' if no arguments are given.

lib/diagnostic.js
```diff
@@ -5,22 +5,36 @@
...
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let array = [];
+  for(let i=0;i<=length-1;i++){
+    array[i] = defaultValue(i);
+  }
```
payne-chris-r, line 36: Think about what defaultValue(i) means--javascript is interpreting this as a function call 'defaultValue' and passing it whatever i is in the for loop as a parameter. This is correct for the 2nd piece, but breaks the first part. You really want to be checking if defaultValue is a function before you invoke it.

payne-chris-r:
Overall, you're very close, and you seem to be getting it. However, because we're giving feedback based on the grunt test, this unfortunately didn't meet the requirements.
Things to work on would be:
- Checking the linter for syntax errors
- Running `grunt test` to check requirements

PR #95, opened 2016-05-23
Arielle js-functions-ins-and-outs-diagnostic response

acobrelo:
Response to diagnostic, mostly completed.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    if (result > arguments[i]) {
+      result = arguments[i];
+    }
+  }
+  return result;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,29 +5,56 @@
...
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
 const newArray = function newArray(length, defaultValue) {
-
+  let array = [defaultValue];
+  for (let i = 1; i < length; i++) {
+    let array += defaultValue;
```
berziiii, line 35: You have already declaired `array`. line 35 is breaking the `grunt test` when we try to run the test script. Be sure to run `grunt test` as the readme states.

lib/diagnostic.js
```diff
@@ -5,29 +5,56 @@
...
+const defaultValue = function (defaultValue) {
+  let result = arguments[0];
+  for (let i = 0; i < arguments.length; i++) {
+    //?
+  }
+}
+const newArray = function newArray(length, defaultValue) {
+  let array = [defaultValue()];
+  for (let i = 1; i < length; i++) {
+    let array += defaultValue(arguments(i));
```
berziiii, line 54: Same as above statement here.

berziiii:
Since the test won't run, we cannot accurately provide feedback. I suggest taking a second look at this dianostic to see where you may have semantic errors. You should also run `grunt nag` to see what is wrong in your code. Moving forward, I highly suggest you run `grunt test` and `grunt nag`; they point out errors.

PR #96, opened 2016-05-23
Erin's Responses

erinsul:


lib/diagnostic.js
```diff
@@ -5,22 +5,36 @@
...
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+    for( let i = 1; i < arguments.length; i++){
+      if(arguments[i] < result){
+        result = arguments[i];
+      }
+    }
+    return result;
```
berziiii, line 14: Passes the test. Nice work

lib/diagnostic.js
```diff
@@ -5,22 +5,36 @@
...
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
-
+  let result = 0;
+    for( let i = 0; i < arguments.length; i++){
+        result = result + arguments[i];
+    }
+    return result;
```
berziiii, line 26: Does not pass the test when no argument is passed into the `sum` function. You may need an `if` statement to check if a argument is passed.

lib/diagnostic.js
```diff
@@ -29,14 +43,30 @@ const newArray = function newArray(length, defaultValue) {
...
+  var array = [];
+  for(i = 0; i<length; i++){
+    if(typeof defaultValue === function){
+    array[i] = defaultValue(i);
+    } else {
+    array[i] = defaultValue;
+    }
+  }
+  return array;
+};
```
berziiii, line 56: This needs to use the `const newArray =` variable. Since you reassigned it, it breaks the tests. Be sure to watch out for reassigned, or trying to, in the future.

berziiii:
1 out of 5. Be sure to run `grunt nag` and `grunt test` as you do these diagnostics. It will tell you semantic errors you have along with what is and is not passing when the tests are run.

PR #98, opened 2016-05-23
Response for Jon Janulewicz

modularjon:
Attempted to get to all questions. No go on 5.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,15 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let result = arguments[0];
 
+  for (let i = 1; i < arguments.length; i++) {
+    if (arguments[i] < result) {
+      result = arguments[i];
+    }
+  }
+
+  return result;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -5,22 +5,46 @@
...
 const min = function min() {
+  let result = arguments[0];
 
+  for (let i = 1; i < arguments.length; i++) {
+    if (arguments[i] < result) {
+      result = arguments[i];
+    }
+  }
+
+  return result;
```
berziiii, line 16: Passes the test. Nice work.

lib/diagnostic.js
```diff
@@ -5,22 +5,46 @@
...
+
+  for (let i = 0; i < arguments.length; i++) {
+    if (arguments) {
+      result += arguments[i];
+    } else {
+      result = undefined;
+    }
+  }
 
+  return result;
```
berziiii, line 34: The `if` statement isn't correct since it doesn't pass the test if there are no arguments.

berziiii:
1 out of 5. I would suggest running the tests via `grunt test` and `grunt nag` as you run through things. This will help guide you towards what needs to be in your code.

PR #99, opened 2016-05-23
Esther Response

esthercordova:


lib/diagnostic.js
```diff
@@ -4,23 +4,46 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
-
+const min = function min(arrayOfNumbers) {
+  let min = function () {
```
berziiii, line 8: You have already assigned the variable `min`. You cannot reuse it  like `let min =` because we already assigned it as a `const`

lib/diagnostic.js
```diff
@@ -4,23 +4,46 @@
...
-
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(arrayOfNumbers) {
+  let result = [];
+  let sum = function() {
```
berziiii, line 24: Same here.

berziiii:
I would suggest trying the diagnostic again but running `grunt test` and  `grunt nag` throughout the diagnostics. The tests wouldn't even run due to have variables attempted to be reassigned.

PR #101, opened 2016-05-23
diagnostic.js submit

mattb5:


lib/diagnostic.js
```diff
@@ -5,30 +5,60 @@
...
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
-
+  let result = 0;
+  for (let i = 0; i < arguments.length; i++)
+  {
+      result = result + arguments[i];
+  }
+  return result;
```
berziiii, line 29: Passes 2/3 tests. be sure to run `grunt test` to see what is not passing and make changes accordingly.

lib/diagnostic.js
```diff
@@ -5,30 +5,60 @@
...
-
-};
+// const newArray = function newArray(length, defaultValue) {
+//   let result = [];
+//   for (let i = 0; i < length; i++)
+//   {
+//     result[i] = defaultValue;
+//   }
+//   return result;
+// };
```
berziiii, line 42: This is where you have to assign `const newArray` and question 4 you have to use the variable in a different way.

berziiii:
By the tests, 1/5 pass. You'll get more practice reading tests as we go but the tests will tell you what passes and what does not. Be sure to check them.

PR #102, opened 2016-05-23
JS Functions ins and outs diagnostic

andreasantarl:
Answers to questions about functions with objects


berziiii:
When tests are run, it won't execute because you have `Unexpected eval or arguments`. This makes me think that you either have a function as an argument that isn't used or you have duplication of variables. Since the tests don't run, we cannot accurately score the diagnostic at this point. Please take a look at the diagnostic and be sure to run tests as you walk through the diagnostic. Let me know if you have questions on this.

PR #106, opened 2016-07-18
Add diagnostic.md week2/day1

NatasaPeic:


RealWeeks:
5/14.  Don't take the number on the score to heart. You did better than your score tells the details of. Your sum and some of your min is correct but your arrays return empty .  Also for your addMethod, it causes the tests to blow up.  Also look what your returning, and where.  Try playing around with this a bit then look at the solution branch for the correct answer.  I think you're closer than you think.

PR #107, opened 2016-07-18
Complete JS Functions Ins/Outs Diagnostic

gabescarbrough:
Questions 1-4 I am confident in my responses. Question 5 I wasn't sure what to do.


RealWeeks:
8/14 Gabe, great job.  Don't let the score mislead you on this, this was good. Looks like you just didn't have time, which was expected here.  Keep up the strong effort.

PR #108, opened 2016-07-18
add answers to diagnostic

dvdpwll:
answerd to diagnostic.


RealWeeks:
9/14.  Great job, looks like you just didn't have enough time which was expected with this. Keep up the great effort.

PR #109, opened 2016-07-18
add solution

EchoXin:
4,4


RealWeeks:
8/14.  Great job, looks like you were a bit rushed at the end.  No worries! We didn't expect people would complete this diagnostic.  Keep up the great work.  Looks like you're returning functions out of your array factory though. Try slowly walking through your code, seeing what it does at each step (what is defaultValue)? Take some time with this then check the solution branch if you need it.

PR #111, opened 2016-07-18
diagnostic 6/18

cdelsoin:
answered some questions


RealWeeks:
3/14.  Don't fret, we didn't expect people to finish this in the allotted time, however this is still not a great score.  I don't even know how accurate it is because you came in late this day. Come in on time please so we better know how to diagnose you and also how the class is doing as a whole.  Refer to the solution branch for the answers.

PR #112, opened 2016-07-18
Sumbitted daily diagnostic

natGalynker:
- had to go back to modify work because I was thinking about the homework rather
  than the in and outs study so wasnt able to answer as many questiosn as I would
  have liked


RealWeeks:
7/14.  I have to give it this score because I can't run the tests and have to test by eye.  Please pay attention to your linter and read your code carefully. It won't run, but it's pretty clear why.  Please feel free to redo and resubmit this

PR #114, opened 2016-07-24
Modified js-functions-ins-and-outs-diagnostic

natGalynker:
-redid the entire diagnostic
-properly commented to the best of my ability
-have a more thorough understanding
- the linter was super finicky about === being correct
- only == would let the grunt test pass for the final question

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -5,7 +5,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result= arguments[0];
+  for (let i = 0; i < arguments.length; i++) {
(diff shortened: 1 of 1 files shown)


laurenfazah:
Hey Natalia - sorry for not seeing this resubmission sooner. There were still syntax issues throughout the diagnostic (like double `return`s in code blocks), but overall, your thought process was spot on. I was really happy to see a lot of commenting and pseudo coding for each question. It appears to have helped break down the larger questions into smaller ones. 

There was one error with question 5. Rather than checking for `null`, you should be checking if object is `undefined`. Your error from running `grunt test` (">> TypeError: Cannot set property 'method' of undefined") should be a hint in the future about that.

Great work, though!

PR #143, opened 2016-10-24
Attempt to complete diagnostic

Bowers96:
Tried answering the questions with the information given.
I'm having a hard time turning the question into code.

Comfort: 3
Clarity: 3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,21 +8,21 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  return Math.min(1, 5);
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(x) {
+  return x = 5 + 2;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray(length, defaultValue) {
 
 };
 
@@ -38,7 +38,7 @@ const newArray = function newArray(/* length, defaultValue */) {
 // If the argument obj is unset, create a new object
 // The function you write should return
 // the object passed in as an argument or the object created
-const addMethod = function addMethod(/* propertyName, method, obj */) {
+const addMethod = function addMethod(propertyName, method, obj) {
 
 };
 


lib/diagnostic.js
```diff
@@ -8,21 +8,21 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  return Math.min(1, 5);
 };
```
neugierige, line 12: Not quite. What you are returning here is the minimum between 1 and 5. That means every time this runs, it will return 1. Here's 1 possible solution:

```
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }

  return result;
};
```

lib/diagnostic.js
```diff
@@ -8,21 +8,21 @@
...
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(x) {
+  return x = 5 + 2;
 };
```
neugierige, line 20: Your function always returns 7. Is that what you intended?? Take a look at this solution:

```
const sum = function sum() {
  let result = 0;
    result += arguments[i];
  }

  return result;
};
```

neugierige: Also, note that the function takes "zero or more arguments." How many does yours take in?

neugierige:
0/4. Hi Brandon, please study the solutions, schedule a 1-1, and review the lesson on ins and outs.

lib/diagnostic.js
```diff
@@ -8,21 +8,21 @@
...
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray(length, defaultValue) {
 
 };
```
neugierige: Please study this solution:
```
    array[i] = defFunc(i);
  }

  return array;
};
```

lib/diagnostic.js
```diff
@@ -38,7 +38,7 @@ const newArray = function newArray(/* length, defaultValue */) {
 // If the argument obj is unset, create a new object
 // The function you write should return
 // the object passed in as an argument or the object created
-const addMethod = function addMethod(/* propertyName, method, obj */) {
+const addMethod = function addMethod(propertyName, method, obj) {
 
 };
```
neugierige, line 43: Please study this solution:

```
  if (!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
};
```

lib/diagnostic.js
```diff
@@ -8,21 +8,21 @@
...
+  return x = 5 + 2;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray(length, defaultValue) {
 
 };
```
neugierige, line 28: Please study this solution for both 3 & 4:

```
    array[i] = defFunc(i);
  }

  return array;
};
```

PR #145, opened 2016-10-24
attempted 3 of 4 of the diagnostic

jbuzzanco:
before time ran out I attempted 3 of 4 of the diagnostic

comfort 2
clarity 3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,7 +7,11 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
+const min = function min(small, medium, large) {
+  small = 1;
+  medium = 2;
+  large = 3;
+  return Math.min(min);
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -7,22 +7,35 @@
...
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
+const min = function min(small, medium, large) {
+  small = 1;
+  medium = 2;
+  large = 3;
+  return Math.min(min);
 
 };
```
neugierige, line 16: Please study the solution and understand why your solution does not work. 

```
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }

  return result;
};
```

lib/diagnostic.js
```diff
@@ -7,22 +7,35 @@
...
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
+const sum = function sum(a, b) {
+  a = 2;
+  b = 2;
+  Math.sum(sum);
```
neugierige, line 25: You need to look up the definition and the syntax of these methods before trying to use them. Please study this solution:

```
const sum = function sum() {
  let result = 0;
    result += arguments[i];
  }

  return result;
};
```

lib/diagnostic.js
```diff
@@ -7,22 +7,35 @@
...
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray(length, defaultValue) {
+  length = [1, 2, 3];
+  defaultValue = 1;
+  for (let i = 0; i < length.length; i++) {
+    return defaultValue;
+  }
+
```
neugierige, line 38: Here's one solution for 3 & 4:

```

    array[i] = defFunc(i);
  }

  return array;
};
```

neugierige:
0/5. Hi Joel, it seems like you are not familiar with the fundamental concepts of ins and outs. You need to go back to that lesson and go over those materials again. Please book a 1-1 with one of us if necessary.

PR #146, opened 2016-10-24
Response

RpRoderick:
diagnostic Mon Oc 24

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -29,13 +29,14 @@ const sum = function () {
     total += amount[i]; // add each element in sum array to total
 
   }
-  return total;
+  return total;               //return total
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function (length, defaultValue) {
+  let newArray = [];
 
 };
 
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,16 +7,29 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
-
+const min = function () {           //the object min is a function with the perameters ()
+  let min = [];                     // min is an array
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -7,22 +7,36 @@
...
+const min = function () {           //the object min is a function with the perameters ()
+  let min = [];                     // min is an array
+  for (i = 0; i < min.length; i++) {    //loop through the array
+    if (value[i] < min[i]){                //if the value of i is less than min??  that
+                                        //  seems wrong
+    }
+  }
+};
+return currentMin;                  //then return the current min
 };
```
neugierige, line 19: A few things to think about:
1) line 12: `min` is an empty array. What would you get by looping through it?
2) line 13: what is `value`? Where/how was it defined? How can you represent the thing you want utilizing the pieces you have?
3) line 18: again, what is `currentMin`? Where/how was it defined?

lib/diagnostic.js
```diff
@@ -7,22 +7,36 @@
...
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
+const sum = function () {
+  let sum = [];
+  for (i = 0; i < sum.length; i++) {  //iterate through the loop of array sum
+    total += amount[i]; // add each element in sum array to total
 
+  }
+  return total;               //return total
```
neugierige, line 32: On line 26, you declared a `const` called sum. On line 27, you declare a `let` also named sum. How is the computer supposed to know which one you want when you call `sum`? If I recall correctly, you've been making the same error in more than 1 diagnostic. Please go back to the fundamentals and review JavaScript variables and functions. Also, try testing your code in node. You will get errors that will tell you what the problem is.

neugierige:
0/5. Hi Ryan, please remember that once you have declared a `const`, you cannot have another variable by the same name. That's like having 2 Bens in the classroom. When someone says "Ben," it's unclear which one is supposed to respond. Humans can understand context, but computers taking everything VERY literally. 

I would encourage you to go over the Fundamentals again, and schedule a 1-1 to review the solutions on this diagnostic.

PR #148, opened 2016-10-24
Adds response to functions-ins-and outs

jdaly04:
Attempts the five questions from functions-ins-and
outs diagnostic.
Comfort: 2
Clarity: 2
I used google, notes, and referenced the solution
branch on the Github repo.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,8 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = Math.min();
+    return result;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,32 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = Math.min();
```
neugierige, line 11: Did you look up the definition/syntax for `Math.min()`? Never copy code blindly. Just Google "Javascript math.min()". 

Please study this solution:

```
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }

  return result;
};
```

neugierige:
1/5. Hi Julia, please make sure you really understand your solution for `sum`. Because you could have used the same strategy for `min`. As for the other questions, please look over the solution branch, and review the last few lessons on JS. Let me know if you need a 1:1.

jdaly04:
Hi Lu,

Could we have a 1-1 today? Not sure how the schedule will work given the
overtime on our last lesson.

Let me know!
Julia

neugierige:
Unfortunately, I'm already booked for today. Jason is out sick, and Ben will be leaving early too. Check with Rachel and Antony. If neither are available, we can try for tomorrow at 4pm.

PR #150, opened 2016-10-24
add responses

treep78:
Comfort: 4
Clarity: 2

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,23 +7,43 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
-
+const min = function min(nums)
+{
+  let result;
+  for(let i in nums)
+  {
+    if (result > nums[i]) {result = nums[i];}
+  }
+  return result;
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(nums)
+{
+  let result = 0;
+  for(let i in nums){result += nums[i];}
+  console.log(i);
+  return result;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
-
+const newArray = function newArray(length, valueFunc)
+{
+  let result = [];
+  if(typeof(valueFunc))
+  {
+    for (let i = 0; i < length; i++) {result[i] = valueFunc(i);}
+  }
+  else
+  {
+    for (let j = 0; j < length; j++) {result[j] = valueFunc;}
+  }
+  return result;
 };
 
 // Question 4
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -7,23 +7,43 @@
...
-
+const min = function min(nums)
+{
+  let result;
+  for(let i in nums)
+  {
+    if (result > nums[i]) {result = nums[i];}
+  }
+  return result;
 };
```
neugierige, line 18: You have the right idea here, but your implementation is not quite right. What's the syntax for for-loop? Please review that. It's very important.

neugierige: Also, your function is not capturing multiple arguments passed in. Do you remember `arguments`? Check out this solution:

```
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }
  return result;
};
```

lib/diagnostic.js
```diff
@@ -7,23 +7,43 @@
...
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(nums)
+{
+  let result = 0;
+  for(let i in nums){result += nums[i];}
+  console.log(i);
+  return result;
 };
```
neugierige, line 30: I'm seeing some of the same error in this code.

neugierige: Here's 1 possible solution:

```
const sum = function sum() {
  let result = 0;
    result += arguments[i];
  }
  return result;
};
```

lib/diagnostic.js
```diff
@@ -7,23 +7,43 @@
...
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
-
+const newArray = function newArray(length, valueFunc)
+{
+  let result = [];
+  if(typeof(valueFunc))
```
neugierige, line 38: What exactly is the syntax for `typeof`? Please take another look. Also, what are you testing for here? What do you _want_ to test for?

lib/diagnostic.js
```diff
@@ -7,23 +7,43 @@
...
+  {
+    for (let i = 0; i < length; i++) {result[i] = valueFunc(i);}
```
neugierige, line 40: Be careful, `[]` != `()`

lib/diagnostic.js
```diff
@@ -7,23 +7,43 @@
...
+  }
+  else
+  {
+    for (let j = 0; j < length; j++) {result[j] = valueFunc;}
```
neugierige, line 44: FYI, because this is a separate loop entirely, you can actually use the variable `i` again.

lib/diagnostic.js
```diff
@@ -7,23 +7,43 @@
...
+  }
+  return result;
 };
```
neugierige, line 47: Here's 1 solution:

```
    array[i] = defFunc(i);
  }
  return array;
};
```

lib/diagnostic.js
```diff
@@ -32,14 +52,20 @@ const newArray = function newArray(/* length, defaultValue */) {
...
-const addMethod = function addMethod(/* propertyName, method, obj */) {
-
+const addMethod = function addMethod(property, method, obj)
+{
+  let affObj = obj || {};
+  affObj.property = function(affObj)
+  {
+    return(affObj);
+  };
 };
```
neugierige, line 69: I think you're on the right track here, but again, the details are off. Notice that you are taking in 3 arguments. Are all 3 used in your code? Take a look at this solution:

```
  if (!obj) {
    obj = {};
  }
  obj[propertyName] = method;
  return obj;
};
```

neugierige:
0/5. Hi Peter, I can tell that you understand which tools to utilize for these problems, but it is also important to make sure that the details are right. For diagnostics, you can always Google for the syntax. And always test your code if you can!

PR #151, opened 2016-10-24
Responded to js-functions-ins-and-outs-diagnostic

Naki21:


lib/diagnostic.js
```diff
@@ -8,6 +8,13 @@
...
 const min = function min() {
+  let min = arguments[0];
+  for (let i = 0; i < arguments.length; i++) {
+    if(min <= arguments[i]){
+      min = arguments[i];
+    }
+    return min;
+  }
 
 };
```
neugierige, line 21: When I tested your code, it only ever returned the 1st value from the arguments. Your if-statement check is not quite right. Check out this solution:

```
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }
  return result;
};
```

lib/diagnostic.js
```diff
@@ -16,14 +23,28 @@ const min = function min() {
...
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray(length, defaultValue) {
+let array = [];
 
+for (let i = 0; i < length; i++) {
+  if (defaultValue()) {
```
neugierige: What does this line of code do? Do you know how to test this out?

lib/diagnostic.js
```diff
@@ -16,14 +23,28 @@ const min = function min() {
...
+    array[i] = defaultValue(defaultValue);
+  } else {
+  array[i] = defaultValue;
+}
+}
+return array;
 };
```
neugierige, line 50: You're very close. Please review this solution:

```
    array[i] = defFunc(i);
  }
  return array;
};
```

lib/diagnostic.js
```diff
@@ -59,14 +61,20 @@ return array;
...
-  obj = {
-    propertyName: method
-  };
+  let obj = new Obj(propertyName, method);
+  return obj;
+
 }
+
+return obj;
 };
```
neugierige, line 78: Your thinking is on the right track, but your syntax is off. Take a look at this solution:

```
  if (!obj) {
    obj = {};
  }
  obj[propertyName] = method;
  return obj;
};
```

neugierige: BTW, when the question prompt says "your function should return..." that means your function needs to have a `return` statement at the end.

neugierige:
1/5. Hi Yiannis, I'm glad to see you coming back to this diagnostic and taking another stab at these questions. Please review my comments and the solutions I posted. Let us know if you want a 1:1.

PR #152, opened 2016-10-24
Response to 10-24-2016 diagnostic

matthewgsommers:
Respond to all questions, was unable to submit question 4's answer.
Felt confident in my other responses. Felt good overall.

Comfort: 3
Clarity: 4
Resources: Used notes from class on ins and outs.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,22 +8,38 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let result = arguments[0]; // if there are any arguments, 'result' equals the first one. But if not, then it returns undefined
 
+  for (let i = 0; i < arguments.length; i++) {
+    if (result > arguments[i]){
+      result = arguments[i];
+    }
+  }
+  return result;
 };
+
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(a,b,c){
+  let result = [];
+  for (let i = 0; i < array.length; i++) {
+    let array[i] = array[()(a + b + c) / 3 )];
+  }
+  return result
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
-
+const newArray = function newArray(length, defaultValue) {
+  let result = [];
+  for (let i = 0; i < array.length; i++) {
+    array[i] = defaultValue[i];
+  }
+  return result;
 };
 
 // Question 4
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,38 @@
...
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(a,b,c){
+  let result = [];
+  for (let i = 0; i < array.length; i++) {
+    let array[i] = array[()(a + b + c) / 3 )];
+  }
+  return result
 };
```
neugierige, line 32: The value we are looking to return here is the sum, not the average. A few things to take note of:
1) line 26: Do you recall the term `arguments`? It's one of the terms that JavaScript provides to us. Because we don't know how many things are going to be passed in, the only we can calculate the sum is to utilize this term.
2) line 27: what is the data type of `result`? you initialized it as an empty array. Should it be an Array?
3) line 28: what is this variable `array`? Where/How did you define it?
4) line 29: this...is a bit of a mess! Recall that the way we access a value inside an array is with `[]`, and the value that goes inside is the index number, which is always an integer, and less than the length of the array itself. What index number is `(a + b + c) / 3`?

neugierige, line 32: Here's 1 possible solution:

```
const sum = function sum() {
  let result = 0;
    result += arguments[i];
  }
  return result;
};
```

lib/diagnostic.js
```diff
@@ -8,22 +8,38 @@
...
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
-
+const newArray = function newArray(length, defaultValue) {
+  let result = [];
+  for (let i = 0; i < array.length; i++) {
+    array[i] = defaultValue[i];
+  }
+  return result;
 };
```
neugierige, line 43: I think you might have misunderstood the question here. Please study this solution (it's for #3 and #4) and make sure you can understand this. 

```
    array[i] = defFunc(i);
  }
  return array;
};
```

lib/diagnostic.js
```diff
@@ -32,16 +48,22 @@ const newArray = function newArray(/* length, defaultValue */) {
...
 // The function you write should return
 // the object passed in as an argument or the object created
-const addMethod = function addMethod(/* propertyName, method, obj */) {
-
+const addMethod = function addMethod(propertyName, method, obj) {
+propertyName[method] = obj;
 };
 
+
+
```
neugierige, line 66: Please study this solution. The syntax for adding a property on an object is `obj[propertyName] = method`.

```
  if (!obj) {
    obj = {};
  }
  obj[propertyName] = method;

  return obj;
};
```

neugierige:
1/5. Hi Matt, I would encourage you to schedule a 1:1 to review this material again, and continue to practice getting used to writing JavaScript syntax for arrays and objects.

PR #154, opened 2016-10-24
Attempt to complete ins and outs diagnostic

akradin:
First two challenges were 5s
Last two were 2s
Used Google, class notes, past pieces of code

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,21 +8,38 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i = 0; i < arguments.length; i++) {
+   if(result > arguments[i]){
+     result = arguments[i];
+   }
+  }
+return result;
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(array) {
+  let result = 0;
+  for (let i = 0; i < arguments.length; i++) {
+    result = arguments[i] + result;
+  }
+return result;
 };
 
 // Question 3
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,21 +8,38 @@
...
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(array) {
+  let result = 0;
+  for (let i = 0; i < arguments.length; i++) {
+    result = arguments[i] + result;
+  }
+return result;
 };
```
neugierige, line 30: FYI, line 27 can also be written as `result += arguments[i]`.

lib/diagnostic.js
```diff
@@ -8,21 +8,38 @@
...
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray(arrayLength, defaultValue) {
+  let result = [];
+
+  for (let i = 0; i < arrayLength.length; i++) {
```
neugierige, line 38: You're a little confused here. What is `arrayLength` supposed to be? It's a length, meaning it is an integer. It is not an actual array itself.

lib/diagnostic.js
```diff
@@ -8,21 +8,38 @@
...
+    result[i] = arrayLength[i];
+  }
+  return result;
+
```
neugierige, line 42: Please study this solution for questions 3 and 4:

```
    array[i] = defFunc(i);
  }
  return array;
};
```

neugierige:
2/5. Good job, Alex. Please study the solutions I included, and also this one for question 5.

```
  if (!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
};
```

PR #155, opened 2016-10-24
Add response for js in out diagnostic.

coffeeisamazing:
Add response for js-functions-ins-and-outs-diagnostic.

Comfort: 3
Clarity: 3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,14 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let result = arguments[0];
 
+  for (let i = 0; i < arguments.length; i++) {
+    if (result > arguments[i]) {
+      result = arguments[i];
+    }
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,45 @@
...
+    if (typeof defaultValue === "function") {
+      result[i] = defaultValue();
+    }
+    else {
+      result[i] = defaultValue;
+    }
+  }
+
+  return result;
 };
```
neugierige, line 50: Very close! For question 4, note that if `defaultValue` is a function, it is supposed to be invoked "with the element's index as an argument." So what you should have in line 42 is actually `result[i] = defaultValue(i)`

Check out this solution:

```
    array[i] = defFunc(i);
  }
  return array;
};
```

neugierige:
4/5. Excellent job, Caleb! Please take a look at my comment for question 4.

PR #156, opened 2016-10-24
Respond to all diagnostic questions

joaufi:
Create functions for each question in the diagnostic

Clarity: 5
Comfort: 4

Google'd like a crazy person

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -63,8 +63,6 @@ const addMethod = function addMethod( propertyName, method, obj ) {
      obj[propertyName] = method;
      return obj;
    }
-   console.log(obj);
-   return obj;
 };
 
 // Used by testing
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,11 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  if (typeof arguments[0] === 'undefined' ){
+    return undefined;
+  } else {
+    return Math.min.apply(null, arguments);
+  }
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,38 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  if (typeof arguments[0] === 'undefined' ){
+    return undefined;
+  } else {
+    return Math.min.apply(null, arguments);
+  }
 };
```
neugierige, line 16: Your solution works exactly as expected. However, you could have accomplished this with simpler code. Your 1st if-clause isn't actually necessary. Take a look at this solution below:

```
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }

  return result;
};
```

lib/diagnostic.js
```diff
@@ -38,8 +54,17 @@ const newArray = function newArray(/* length, defaultValue */) {
...
+     let obj = {};
+     obj[propertyName] = method;
+     return obj;
+   } else {
+     obj[propertyName] = method;
+     return obj;
+   }
+   console.log(obj);
+   return obj;
 };
```
neugierige, line 66: Did you have a chance to test this code with an obj that does not yet exist? I got a reference error saying that the thing I passed in as the `obj` argument is not defined. Take a look at this solution:

```
  if (!obj) {
    obj = {};
  }

  obj[propertyName] = method;

  return obj;
};
```

neugierige: Also, notice that both clauses of your if-statement are running some of the same code. That should raise a red flag. It signals that they don't actually need to be inside the if-statement. Make sense?

neugierige:
4/5. Great job, Johnathan! Please take a look at the solution for question 5. See what you can change in your code to make it work as expected.

neugierige:
5/5. Johnathan, please ignore my previous comment. Your solution for question 5 does work as expected, although you can make it slightly more efficient. Excellent job!

PR #159, opened 2016-10-24
Respond to diagnostic

rickdean:
Respond to diagnostic

Comfort: 2
Clarity: 3

I used Google and my class notes to help with this assessment.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,21 +8,34 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let end = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    if (arguments[i] < end) {
+      end = arguments[i];
+    }
+  }
+  console.log(end);
 };
+
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
+  let end = 0;
+    for (let i = 0; i < arguments.length; i++) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,21 +8,34 @@
...
 const min = function min() {
-
+  let end = arguments[0];
+  for (let i = 1; i < arguments.length; i++) {
+    if (arguments[i] < end) {
+      end = arguments[i];
+    }
+  }
+  console.log(end);
 };
```
BenGitsCode, line 19:
``` js
const min = function min() {
  let result = arguments[0];
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }

  return result;
};
```

this is really close, i think using different naming conventions confused you a bit. This is technically right but you want to use const (this shouldn't change) and use conventions like result instead of end. It's right, I just wanted you to know the proper naming convention.

lib/diagnostic.js
```diff
@@ -38,10 +51,27 @@ const newArray = function newArray(/* length, defaultValue */) {
 // If the argument obj is unset, create a new object
 // The function you write should return
 // the object passed in as an argument or the object created
-const addMethod = function addMethod(/* propertyName, method, obj */) {
+const addMethod = function addMethod(propertyName, method, obj) {
+  if obj  = {} {
+    obj[propertyName] = method;
+  }
+};
```
BenGitsCode, line 59:
``` js
    array[i] = defFunc(i);
  }

  return array;
};
```

BenGitsCode (commented):
2/5

PR #160, opened 2016-10-24
response to diagnostic

kfili:
functions ins and outs

comfort: 2
clarity:3


BenGitsCode (commented):
2/5

Your code here is accurate but the convention is to use result where you used x. It's right and it works, I just wanted you to know that.

PR #161, opened 2016-11-21
better than last week - still incomplete

NewBeg71:
Newbeg71 - Week 3 - Monday - Morning Diagnostic - js-ins-out 

Comfort: 2
Complete: 2

w3, "Javascript is sexy", stackoverflow(not so much this one


estermer:
Looks like you didn't find the arguments object which is accessing the arguments passed to a function not already specified with parameters.  If you have any questions on this you can slack me.

PR #165, opened 2016-11-21
neilburd week 02 day 01 DIAGNOSTIC

neilburd:
comfort: 3
completeness: 3
comment:
I feel as though I understand the basic concepts but still don't get the adding methods.


estermer:
In these functions we were looking for the arguments object to be used which is a dynamic object of arguments that were passed to the function that were not expected with predefined parameters. Nice attempt on the ones you completed. In the future we will get some code challenges where you can see you output to really hone in you skills.

PR #166, opened 2016-11-21
murphypicard Ara diagnostic week 2 day 1

MurphyPicard:
comfort 3
completeness 3

Kind of uncomfortable.  It takes me a while to think what I want to put into the function and what I want to get out.  I didn't attempt the 4th question.


estermer:
In this diagnostic you were going to access the arguments object which is a dynamic list of arguments passed to the function that were not specified with parameters already. In the future we will do code challenges where they are more isolated and you can view your output better to see whats working and what isnt.

PR #167, opened 2016-11-21
response

cliftonh02:
Comfort: 0
Clarity: 0

I used Google and youtube but its clear that i don't have an understanding of how to write functions. I really want to grasp this concept however I'm struggling with the "grammer" of a function sentence  and how to translate in the editor if that makes sense. I will need to refer to class notes and do more research to accomplish this.


estermer:
Slack me and i can walk you through this a little. -Eric

PR #171, opened 2016-11-21
week 2 day 1 diagnostic @ahnuce

ahnuce:
Comfort level: 2
completeness: 2
The setup of the functions are giving me trouble.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,22 +8,34 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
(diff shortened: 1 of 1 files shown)


JosephinaTavera:
✅ 

Quick notes:  

(1) Re-read the following document 

https://github.com/ga-wdi-boston/js-functions-ins-and-outs


(2) Review the solution document for the answer to the other two questions.  

https://github.com/ga-wdi-boston/js-functions-ins-and-outs-diagnostic/blob/solution/lib/diagnostic.js

Then let me know if you still have questions.

PR #172, opened 2016-11-21
answer 3 out the 5 questions

RFSync:


JosephinaTavera:
✅ 

Quick notes:  

(1) Re-read the following document 

https://github.com/ga-wdi-boston/js-functions-ins-and-outs


(2) Review the solution document for the answer to the other two questions.  

https://github.com/ga-wdi-boston/js-functions-ins-and-outs-diagnostic/blob/solution/lib/diagnostic.js

Then let me know if you still have questions.

PR #173, opened 2016-11-21
mwb86 Week 02 Day 01 diagnostic

mwb86:
comfort 2
completeness 2


estermer:
Good attempts, remember when accessing the arguments object it is plural `arguments`.  We will get into code challenges later on so you can run your code and see what works and what doesn't to really get the code working.

PR #174, opened 2016-11-21
Week 02 Day 01 Diagnostic

DomreyCS:
Comfort: 1
Completeness: 1

Comment: I had a hard time with this diagnostics, I definitely need more time reviewing functions and arrays.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,8 +8,16 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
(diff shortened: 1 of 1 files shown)


estermer:
I can tell you definitely had a hard time. Slack me if you have any specific questions.

Also in the question you attempted, remember math is all caps `MATH.min` but for the most part you were really close. It looked like you understood the arguments object which is good. It is just an object that access the arguments that are passed that were not already anticipated with parameters in the function.

PR #176, opened 2016-11-21
answered questions

missjuliekhun:
Fist to five: 2 
I used Google, and the readings assigned as well as tested in the console.


estermer:
definitely need to look up the `arguments` object in js. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

PR #185, opened 2017-01-18
Respond to diagnostic

samordonez:
-created functions asked for in lib/diagnostic.js

Comfort: 4
Clarity: 4

I used google to find a way to check if
a variable was a function, and found an
answer on Stack Exchange.


bernardlee:
5/5

PR #187, opened 2017-01-18
Add answers to ins and outs diagnostic

akgoode:
Comfort: 5
Clarity: 5

Needed to refer to old js-objects repositories to figure out the best way to set propertynames.

My test for question 3 worked until I added the code for question 4.  Then the test for 3 failed but the code for 4 worked.
If you comment out question 3's method and uncomment 4's it will work.  If you uncomment both then question 5 will fail.


bernardlee:
5/5

PR #192, opened 2017-01-18
Diagnostic - JS functions ins and outs

zchenbonterra:
Comfort: 3
Clarity: 3

Used google for question 4, but couldn't figure a way out to solve it. Please advise!!!


bernardlee:
4/5.  For question 4 you want to use the ```typeof``` unary operator.  ```typeof``` will return the type of the input.  For example, ```typeof 3``` will return ```'number'``` and ```typeof 'hello'``` will return ```'string'```.

PR #197, opened 2017-01-18
ins and outs diagnostic

cshea93:
comfort: 2

clarity: 3

Personally, while yesterdays lesson was really helpful in finally getting a grasp for functions, I was unable to really absorb the information while doing homework/outcomes in the afternoon. Was never able to complete my lab

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,7 +7,18 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
+const min = function min(numOne, numTwo) {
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -32,6 +52,13 @@ const newArray = function newArray(/* length, defaultValue */) {
 // to the result of invoking defaultValue with the
 // element's index as an argument
 
+const newArray = function newArray(){
```
bernardlee, line 55: ```newArray``` was already defined earlier by using ```const``` so trying to redefine or reassign it will result in an error.

bernardlee:

PR #198, opened 2017-01-18
Update answers

aneela06:
Comfort: 3 -- I still need more work with functions.
Clarity: 4 

I used my class notes to help with this diagnostic and I tried to do it on my own to see if I could get it or not.  I think understanding where to begin is hard for me.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i = 0; i < arguments.length; i++){
+    if (result[i] < arguments[i]){
+    return result[i];
+    }
+    return result;
+  }
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,36 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments[0];
+  for (let i = 0; i < arguments.length; i++){
+    if (result[i] < arguments[i]){
+    return result[i];
```
bernardlee, line 14: There shouldn't be a ```return``` here since it will exit the loop and function.

lib/diagnostic.js
```diff
@@ -8,22 +8,36 @@
...
+    }
+    return result;
```
bernardlee, line 16: This ```return``` will exit the loop and function after one iteration.

bernardlee:
1/5

PR #199, opened 2017-01-18
diagnostics response

moshiko1988:
i was struggling with understanding the questions

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,11 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments ;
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,35 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = arguments ;
+  for (let i = 0; i < arguments.length; i++) {
+    result = arguments[i];
+  }
+return result;
```
bernardlee, line 15: This function and ```sum()``` will only return the argument if there is one argument, and the last argument if there is more than one argument.

bernardlee:

PR #202, opened 2017-01-18
Max's Response

doremaxime:
comfort: 3
clarity: 3

Checked out http://stackoverflow.com/questions/9005778/javascript-push-array-onto-array-with-for-loop

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,7 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  //returns the minimum amount of arguments? or the smallest interger?
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,30 +8,50 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  //returns the minimum amount of arguments? or the smallest interger?
```
bernardlee, line 11: Number.  The question states, "assume that all the argument are numbers".

lib/diagnostic.js
```diff
@@ -8,30 +8,50 @@
...
+sum([1, 2, 3, 4]);
+
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray(length, defaultValue) {
+  let array = [];
+  for (let i = 0; i < length; i++) {
+    array.push[i](defaultValue);
```
bernardlee, line 35: ```push``` is an array method that adds an element onto the end of the array.  This should be ```array.push(defaultValue);```.

lib/diagnostic.js
```diff
@@ -8,30 +8,50 @@
...
+newArray(6);
+
+
 // Question 4
 // Add code to newArray so that if defaultValue is a
 // function then the elements in the array will be set
 // to the result of invoking defaultValue with the
 // element's index as an argument
 
+// From what i understand you would like us to add a callback?
```
bernardlee, line 52: We want you to determine whether defaultValue is a function or not.

bernardlee:
1/5

PR #203, opened 2017-01-18
js diagnostic jan 18

ggabrakadabra:
-completed diagnostic questions

Comfort:2
Clarity:2

Can you send me tips/tricks to get better at this? Or even other tutorials? I would like to be able to complete a diagnostic

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,11 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = [];
+  for (let i = 0; i < min.length; i++) {
+    result[i] = min[i];
+  }
+  return result;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,33 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+  let result = [];
+  for (let i = 0; i < min.length; i++) {
```
bernardlee, line 12: ```min``` is the name of the function and its length is 0.  This condition evaluates to ```0 < 0```, which is false, so the loop never runs.

lib/diagnostic.js
```diff
@@ -8,22 +8,33 @@
...
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
-
+  let result = [];
+  for (let i = 0; i < sum.length; i++) {
```
bernardlee, line 24: This has the same issue as line 12.

bernardlee:
0/5.  Feel free to ask for help if you are having trouble.  I will write and add a solution to the code channel later on for everyone to look at.  I can't think of any tips or tricks off of the top of my head, but you can try the free JavaScript tutorial at https://www.codecademy.com.

PR #205, opened 2017-01-18
answers to diagnostic

Soboscott:
still intimidated by writing code from scratch

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,7 +7,10 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
+const min = function min(arguments) {
+  result = 0
+  for (let i =0; i < arguments.length; i++ )
+  result +=arguments[i];
 
 };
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -7,22 +7,30 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min() {
+const min = function min(arguments) {
```
bernardlee, line 10: This will throw an error because ```arguments``` is a reserved keyword.

lib/diagnostic.js
```diff
@@ -7,22 +7,30 @@
...
 
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
-
+const sum = function sum(arguments) {
```
bernardlee, line 21: This will throw an error because ```arguments``` is a reserved keyword.

bernardlee:

PR #207, opened 2017-01-18
Complete  diagnostic

ddobson:
Answers added in lib/diagnostic.js

comfort: 3/5

clarity: 3/5

I had trouble understanding directions in questions 4 and 5.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,14 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let mini = arguments[0];
 
+  for (let i = 0; i < arguments.lenth; i++) {
+    if (mini > arguments[i]) {
+      mini = arguments[i];
+    }
+  }
+  return mini;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,41 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let mini = arguments[0];
 
+  for (let i = 0; i < arguments.lenth; i++) {
```
bernardlee, line 13: This typo causes the loop to never run.  As it is, it evaluates ```i < arguments.lenth;``` as ```0 < undefined```, which evaluates to false.

lib/diagnostic.js
```diff
@@ -38,8 +57,12 @@ const newArray = function newArray(/* length, defaultValue */) {
 // If the argument obj is unset, create a new object
 // The function you write should return
 // the object passed in as an argument or the object created
-const addMethod = function addMethod(/* propertyName, method, obj */) {
+const addMethod = function addMethod(propertyName, method, obj) {
+  if (obj === undefined) {
+    obj = {};
+  }
```
bernardlee, line 64: We want to set a property called ```propertyName``` on ```obj```  with ```method``` as its value.

bernardlee:
2/5. For question 4, we want you to modify ```newArray()``` so that if ```defaultValue``` is a function, then it sets each element in the array to the value of ```defaultValue()``` with the element's index as the argument.

PR #208, opened 2017-01-18
Diagnostic ins and outs

laurpaik-zz:
--ran grunt test midway through and it approved what I had,
--but now it's not approving anything
--I didn't change what I had earlier so idk
--comfort: 3, stumbled some on the last few questions
--clarity: 3,

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,14 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let lowNum = arguments[0];
+  for (let i = 0; i < arguments.length; i++) {
+    if (lowNum > arguments[i]) {
+      lowNum = arguments[i];
+    }
+  }
 
+  return lowNum;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,40 +8,71 @@
...
 // the object passed in as an argument or the object created
-const addMethod = function addMethod(/* propertyName, method, obj */) {
 
+let hat = {
+  name: 'hat',
+  color: 'orange',
 };
 
+const addMethod = function addMethod(propertyName, method, obj) {
+  propertyName[method] = obj;
```
bernardlee, line 71: This line breaks the test because it's trying to set a property called ```method``` onto the string ```propertyName``` with ```obj``` as the value.  The correct line is ```obj[propertyName] = method;``` if ```obj``` is an object.

lib/diagnostic.js
```diff
@@ -8,40 +8,71 @@
...
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
+  let nums = [];
+  let result = 0;
+  for (let i = 0; i < nums.length; i++) {
```
bernardlee, line 28: ```nums``` is an empty array that you're iterating over, so this loop never runs.

lib/diagnostic.js
```diff
@@ -8,40 +8,71 @@
...
 
+  return result;
 };
 
 // Question 4
 // Add code to newArray so that if defaultValue is a
 // function then the elements in the array will be set
 // to the result of invoking defaultValue with the
 // element's index as an argument
+const defaultValue = function () {
```
bernardlee, line 52: We want you to modify ```newArray()``` here, not write a new function.

bernardlee:
2/5.  Feel free to ask for help if you are having trouble.  I will write and add a solution to the code channel later on for everyone to look at.

PR #209, opened 2017-01-18
Create functions

skylarkJ:
5 functions

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,14 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let minNum = arguments[0];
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,22 +8,47 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let minNum = arguments[0];
 
+  for (let i = 0; arguments[i].length; i++) {
```
bernardlee, line 13: Since ```arguments``` is expected to be empty or an array containing at least one number, ```arguments[i].length``` is attempting to find a ```length``` property on a number and it will evaluate to ```undefined```.  Thus, this loop will never run.

bernardlee:

PR #210, opened 2017-01-18
Diagnostic response 1.18

K0iram:


lib/diagnostic.js
```diff
@@ -8,21 +8,40 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
```
bernardlee, line 10: This function will only return the first argument (or ```undefined``` if there are no arguments).

lib/diagnostic.js
```diff
@@ -8,21 +8,40 @@
...
+
+  return result;
 
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray() {
```
bernardlee, line 38: ```length``` and ```defaultValue``` should be function parameters.

bernardlee:
1/5.  Feel free to ask for help if you are having trouble.  I will write and add a solution to the code channel later on for everyone to look at.

PR #211, opened 2017-01-18
pushing diagnostic

novakw267:
Terminal was giving me trouble trying to push this.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,21 +8,40 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let lowNum = arguments[0];
 
+  for (let i = 0; i < arguments.length; i--) {
+    if (lowNum < arguments[i]) {
+      lowNum = arguments[i];
+    }
+  }
+
+  return lowNum;
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
+const sum = function sum(num1, num2) {
+  let sumOf = num1 + num2;
 
+  return sumOf;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,21 +8,40 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
+  let lowNum = arguments[0];
 
+  for (let i = 0; i < arguments.length; i--) {
```
bernardlee, line 13: This decrements ```i``` after each iteration so that it'll always be less than ```arguments.length``` if any arguments are passed into the function, causing this to become an infinite loop.

lib/diagnostic.js
```diff
@@ -8,21 +8,40 @@
...
+
+  return lowNum;
 };
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum() {
+const sum = function sum(num1, num2) {
```
bernardlee, line 26: This should have been left as is, since there might no, one, or more than two arguments.

lib/diagnostic.js
```diff
@@ -8,21 +8,40 @@
...
+  let sumOf = num1 + num2;
 
+  return sumOf;
 };
 
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray(/* length, defaultValue */) {
+const newArray = function newArray(/* length, defaultValue */arr, i) {
```
bernardlee, line 35: ```length``` and ```defaultValue``` should be left as the parameters of this function.

bernardlee:

lib/diagnostic.js
novakw267: The explanation of for loops today during the lesson helped a little bit too. I second guessed myself and assumed it had to be different.

PR #212, opened 2017-01-18
Complete

AshlenPrice:
Comfort 2.5
clarity 2.5

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
-
+let result = arguments[0];
+for (let i = 1; i < arguments.length; i++) {
+if (arguments[i] < result){
+result = arguments [i];
+}
+}
+return result;
 };
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -8,15 +8,25 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min() {
```
bernardlee, line 10: Structure your code using whitespace and indentation to increase readability.  While this function is correct, it should be written for readability as well.
```javascript
const min = function min() {
  let result = arguments[0];
  
    if (arguments[i] < result) {
      result = arguments[i];
    }
  }
  
  return result;
};
```

lib/diagnostic.js
```diff
@@ -8,15 +8,25 @@
...
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
 const sum = function sum() {
-
+let result = 0;
+  for (let i = 0; i < arguments.length; i++){
+result += arguments;
```
bernardlee, line 27: This should be ```arguments[i]``` since you're looping through ```arguments``` and you want to add the element at index ```i```.

bernardlee:
1/5.  Feel free to ask for help if you are having trouble.  I will write and add a solution to the code channel later on for everyone to look at.

PR #218, opened 2017-03-28
Response to diagnostic

rrdaniels85:
Comfort: 5/5
Clarity: 4/5

I used class notes.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,23 +8,38 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
-
+  let result = arguments[0]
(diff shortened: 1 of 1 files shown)


rrdaniels85:
Side note: I commented out my function for #3 because I wasn't sure if I could have two functions with the same name on the page.  But the commented out section for # 3 is my answer for that question

jordanallain:
4/5

Great job Rachel, you were like 99% of the way there. There was a gotcha with the `min()` function that if it didn't receive any arguments it should produce `0` but yours returned `undefined`. Otherwise you nailed it though! Keep up the great work.

🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #219, opened 2017-03-28
Functions Ins and Outs

antleo1995:
Got to the 4th question and was getting "defaultValue not a function"
Wasn't able to suss out what was wrong with my code in the allotted time
but I feel like I wasn't far off and could easily figure out what was
wrong given just a little longer. Will review during workshop time.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,12 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
-
+  let result = arguments[0]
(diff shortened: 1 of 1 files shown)


jordanallain:
2/5

Hey Tony this was definitely a tough diagnostic. You pretty much got the first one right, but there was a gotcha where it was supposed to return `0` if you didn't pass it any arguments and yours returned undefined.

You got questions 2 and 3 correct so great job there. You can review the solution branch if you want and if something about it doesn't make sense please let me know!

🇯🇴 🇯🇴

PR #221, opened 2017-03-28
Committing Ins and Outs js diagnostic

deeforte:
committing the March 28th diagnostic
clarity: 4
comfort: 4

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
-
+  let result = arguments[0];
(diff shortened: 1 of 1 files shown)


jordanallain:
0/5

Hey Dee Dee so you had some syntax errors in your code that prevented the tests from passing. In your for loops you have `i = 0` when you should have `let i = 0`. After I fixed that you passed the tests for `min` and `sum` so good job! Make sure you are watching your linter and you can run `grunt make-standard` from the command line which will either fix syntax or formatting issues, or let you know that you have a big one in your code that it can't fix automatically.

As for the last 3 the working code is on the solution branch  so I would take a look at those and see if they make sense to you. If not we can go over them together so feel free to reach out.

PR #222, opened 2017-03-28
Functions Ins and outs assignment

glasneed:
Ran out of time for last question.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,23 +7,46 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min () {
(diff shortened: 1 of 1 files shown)


jordanallain:
0/5

Hey Conor seems like you struggled with this. Take a look at the solution and we can go over anything that isn't clear to you. I also noticed you were starting your for loops like:

`for(let int i = 0`

Not sure where you picked up that `int` from but we don't assign a type with javascript when declaring a variable. `let i = 0` is all we need.

glasneed:
Thanks Jordan. Yeah, I’m lost on callbacks. I’ll try to look at the solution tonight (assuming we don’t have a lot of homework) to see if any lights go off. If not, maybe we can meet tomorrow.

Thanks

PR #223, opened 2017-03-28
completed the js-func-ins-and-outs diacnostic

gastevend:
under the new response branch


jordanallain:
1/5

Hey Steve this was a tough one for sure. You got the `sum` function working, and you were super close on the `min` function. You set up the number as a constant though so then it won't allow you to change it later in the function. The last three were a bit tougher so if you need to go over the solutions and reach out to me please feel free!

🇯🇴

PR #224, opened 2017-03-28
JS Functions-Ins/Outs Diagnostic

cpm5041:
Comfort: 2
Clarity: 3

I used our training exercises primarily for this assignment. I
felt ok completing the simpler ones but not as much with the callback
functions.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,12 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
-
+  let result = arguments[0]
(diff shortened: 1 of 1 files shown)


jordanallain:
1/5

This was a tough one Chris. You got `min` right, and you were REALLY close to getting `sum`. There was a gotcha where if the function wasn't passed any arguments it returned `0`, but yours returned `undefined`. So good job on those two.

The other three were more difficult, I suggest you go over the solution branch and let me know if anything in there doesn't make sense to you.

🇯🇴

PR #226, opened 2017-03-28
Submitting Diagnostic

sjruby:
Comfort 3/5
Clarity 4/5

I think I could have finished these, and probably will during the
the workshop time.  I'm a little confused as to why the min
function isn't working but just need to bang my head against it.
Didn't really get time to test/fix the object method function, but
I think I'm heading in the right direction.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,17 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
-
+  let result = 0
+  if (arguments.length === 1){
+    result = arguments[0]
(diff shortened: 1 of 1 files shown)


jordanallain:
2/5

Hey Sam, thanks for the great feedback. You were super close on `min`, the gotcha was that if it was passed zero arguments it should return `undefined`, otherwise you would have nailed that one too.

You were able to return an array of specified length full of default values, but you got a little hung up on using a function to produce the default values. The `addMethod` was a tough one so check out the solution code and let us know if you need more help.

🇯🇴 🇯🇴

PR #228, opened 2017-03-28
Responses: JS Functions Ins/Outs

naributler:
Comfort: 2/3
Clarity: 3

Ran out of time to do part of the diagnostic. Used class notes to complete the exercise.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,24 +7,47 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min () {
-
+const min = function min (array) {
+  let result = array[0]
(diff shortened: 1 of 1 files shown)


jordanallain:
3/5

Hey Nari, so for `min` and `sum` your functions work but you made them accept arrays as an argument when they were supposed to take multiple numbers as arguments. They worked once I figured that out but they wouldn't have passed the tests we set up because they were expecting arrays instead of numbers.

You got the function for making an array of specified length with a default value, and you were really close on the other one that took a function to produce the default value, but instead of making an array of the return value of the function, it made an array of the same function for every default value. You can check out the solution code for the last one if you want otherwise you're doing great.

🇯🇴 🇯🇴 🇯🇴

PR #230, opened 2017-03-28
Response

rlarsonil71:
Comfort :4 (questions 1-3)
Comfort: 3 (questions 4-5)
Clarity: 3

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,15 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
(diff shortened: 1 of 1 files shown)


jordanallain:
2/5

Hey Ron good job here. You got `min` and you would have got `sum` but it was supposed to return 0 if you didn't pass it any arguments and yours returns undefined. You were able to return an array of default values but not questions 4 where you use a callback function to produce the default value.

For the last one you were SUPER close but you check if the object is null when you should check if it is undefined. Once I changed that it worked as expected so great work.

🇯🇴 🇯🇴

PR #231, opened 2017-03-28
Diagnostic

brianedmunds:
My diagnostic answers.

Comfort: 2
Clarity: 4

I used Google and my notes from previous lessons.


jordanallain:
1/5

Hey Brian you got `sum` but when you were supposed to write a `min` function you wrote a `max` function so not sure what happened there. I think after the extra lesson we did on callback functions that you probably have a better grip on what's going on but if you need extra help let us know.

🇯🇴

PR #233, opened 2017-03-28
Fallon diagnostic

jonifallon:
Fail on my part.  I do not understand what we did yesterday and had no
time to review last night.  I know many answers
could likely be found here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,21 +8,26 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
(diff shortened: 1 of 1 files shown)


jordanallain:
0/5

Hey Joni this was a tough one. I think after we had the bonus lesson on callbacks you would have done better on this. The solution branch is there for review but if something is still a little foggy don't hesitate to set up a 1 on 1 through Chris.

PR #234, opened 2017-03-28
Diagnostic for JavaScript Functions - "Ins & Outs"

octater:
Comfort: 3
Clarity: 3

I used my class notes to help with this diagnostic.


jordanallain:
2/5

This was a tough one Terry and I bet you'd do better on it after the extra lesson we did on callbacks. Take a look at the solution or try these again and we can always set up another time to talk about anything here if it doesn't make sense still after that.

🇯🇴 🇯🇴

PR #236, opened 2017-03-28
finished exercisises almost

MattKwonDo:
3 comfort and clarity


jordanallain:
0/5

Hey Matt you almost got `sum` and `min` there were just a couple things wrong there. I suspect that after our second lesson on callbacks we did on Thursday that you probably feel a little better about some of these. If you need to go over any of this again let us know before we move forward.

PR #237, opened 2017-03-28
Answers to JS functions in and out diagnostic

alzateja:
Comfort: 3
Clarity:3
I did not really understand the last question and I ran out
of time.  I am still having trouble passing functions in as
arguements.  4 and 5 were very difficult for me.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,13 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
-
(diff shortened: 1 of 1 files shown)


jordanallain:
4/5

Great job Jayson! This was a tough diagnostic and you nailed the four that you completed. I suspect that after our bonus callback lesson you might feel better about the one you didn't get but if you want to go over any of this again reach out to Chris and he can set you up with a 1 on 1 with one of us.

PR #238, opened 2017-03-28
Add response

khan-ngo:
3/5

I relied heavily on notes. I understand concept but cannot create from scratch


jordanallain:
1/5

Hey Khanh you only got one of these but I bet you'd do better after we had the extra lesson on callbacks. If you still need a little help with any of this after looking at the solution branch let Chris know and he'll set you up with a 1 on 1 with someone.

🇯🇴

PR #239, opened 2017-03-28
Ins and Outs

Ryaong792:
was able to get 1 and 2 but after was not sure how to start.  I
think i might need more help understanding the questions being ask.
fist of 5 i say i give it a 1 or 2 max.


jordanallain:
2/5

Hey Ryan you got a couple of these and were close on another one. I bet you would do even better today after going through the second callback lesson we did. If you still need any help though reach out to Chris and he will set you up with a 1 on 1 with one of us.

🇯🇴 🇯🇴

PR #241, opened 2017-03-28
response

donpowers:


jordanallain:
3/5

Awesome job Don, you did great on three of these. Check out the solution if you need to and let us know if you want to set up a 1 on 1 to go over anything in here that might not make sense.

🇯🇴 🇯🇴 🇯🇴

PR #242, opened 2017-03-28
Amanda's JS Functions Ins & Out

aeabbott:
First three questions went well! Struggled with the number four an couldn't make it
to number five


jordanallain:
2/5

Hey Amanda this was in a little after 9:30 so try to get the PR open before then even if you haven't finished what you are working on. You got `min` and `sum` so great job with those! I bet you'd do even better after our second lesson on callbacks. Talk to Chris about setting up a 1 on 1 with one of us though if you still need some help.

🇯🇴 🇯🇴

PR #244, opened 2017-04-04
Bkim js-functions first changes

TronB:
Comfort: 2
Clarity: 2-3
Works, but I don't think it's the answer that a consultant is looking for


benjimelito:
You had the right idea for both of these functions, but each of them should be able to take an unlimited number of arguments (rather than just two). This can be accomplished using the `arguments` feature within your function. 1/5

PR #246, opened 2017-04-04
Responses for js functions diagnostic

kellye2010:
Comfort: 3
Clarity: 3

I used mostly my class notes to help with this diagnostic. I pushed early to ensure enough time, but sacrificed missing answering a few problems. I believe I could've completed the task with more time, hence the 3, but I do need more practice on this topic over all.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -8,7 +8,14 @@
 // and returns the minimum from that set
 // assume that all the argument are numbers
 const min = function min () {
+  let result = arguments[0]
 
+  for (let i = 0, min <= arguments.length; i > min; i++) {
+    const current = arguments[i]
+    result = current < result ? current : result
+
+  }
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -22,8 +29,9 @@ const sum = function sum () {
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray (/* length, defaultValue */) {
-
+const newArray = function newArray (length, defaultValue) {
+    let defaultValue = 0;
+    return newArray.length;
```
benjimelito, line 34: Yup, but you need some way to set the items in the array.

benjimelito:
Nice answer to question one. Don't worry about not completing the diagnostic, you'll get faster with practice. 1/5

PR #247, opened 2017-04-04
Addeed my responses

ghost:
comfort : 2
clarity: 2

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,23 +7,38 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min () {
+const min = function min (a) {
+  let result = aguments[0]
 
+  for(let i = 0; i <aruments.length; i++){
+    if result < arguments[i]) {
+      result =arguments[i];
+    }
+  }
+  return result;
 }
 
 // Question 2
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -7,23 +7,38 @@
...
 }
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum () {
-
+const sum = function sum (num1, num2) {
+  return  num1 + num2
```
benjimelito, line 26: Yes, but the function should be able to accept an unlimited number of arguments. Similar to the function above ^^

lib/diagnostic.js
```diff
@@ -7,23 +7,38 @@
...
 // Question 3
 // Write a function that returns an array of the specified length
 // with all the elements set to defaultValue
-const newArray = function newArray (/* length, defaultValue */) {
+const newArray = function newArray (length, defaultValue) {
+
+    let result = [];
+
+    for (let i = 0; i < arguments.length; i++) {
+      result[i] = arguments[i];
```
benjimelito, line 37: You want to set these values to the `defaultValue` argument that was passed in

benjimelito:
Nice answer for #1, we'll go over some of this during workshop later today. 2/5

PR #248, opened 2017-04-04
Answering questions, number five seems hard

Anddrew914:


benjimelito:
Lookin good. Don't overthink number 5, it's basically just asking you to add a method to an object, and if there is no object passed in, create one yourself. 4/5

PR #250, opened 2017-04-04
JS Ins and Outs Diagnostic

bcem2004:


lib/diagnostic.js
```diff
@@ -7,31 +7,43 @@
...
 
 }
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum () {
+const sum = function(num1, num2) {
+  return num1 + num2
```
benjimelito, line 22: Yup, but this function should be able to take an unlimited number of arguments. Using the `arguments` keyword inside your function would be a helpful way to get an array of all the arguments that were passed in.

benjimelito:
Nice attempt, we'll go over some of these concepts more in depth during workshop today if you have any lingering questions. 1/5

PR #251, opened 2017-04-04
response

eliottenos:
Not confident but I know more than what is reflected here!

Comfort: 1
Clarity: 1

used stackoverflow, w3schools, notes and class material

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,22 +7,39 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min () {
+const min = function min (array) {
+  let result = array
 
+  for (let i = 1, max = array.length; i < max; i++) {
+    const current = array[i]
+    result = current > result ? current : result
+  }
+  return min()
 }
+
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -7,22 +7,39 @@
...
 // assume that all the argument are numbers
-const min = function min () {
+const min = function min (array) {
+  let result = array
 
+  for (let i = 1, max = array.length; i < max; i++) {
+    const current = array[i]
+    result = current > result ? current : result
+  }
+  return min()
```
benjimelito, line 17: `min` is the name of your function, I think you actually want to return `result` here!

lib/diagnostic.js
```diff
@@ -7,22 +7,39 @@
...
+
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum () {
+const sum = [1, 2, 3].reduce(add, 0);
 
+function add(a, b) {
+  return a + b;
```
benjimelito, line 30: Yup, but you want to allow your function to take more than just two arguments

benjimelito:
No worries, we'll go over some of this later today in workshop. 1/5

PR #254, opened 2017-04-04
Answer JS Functions ins and outs diagnostic

nriser:
Could not finish the diagnostic. Could not solve the problems now,
but will review them later and will solve them.

### DIFF
--- a/lib/diagnostic.js
+++ b/lib/diagnostic.js
@@ -7,17 +7,32 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min () {
 
+num = []
+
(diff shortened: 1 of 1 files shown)


lib/diagnostic.js
```diff
@@ -7,45 +7,90 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min () {
 
+const min = function min () {
+  return Math.min.apply(null, arguments)
```
benjimelito, line 12: Nice clean solution!

benjimelito:
Your answers look really good to this, I think you only failed to complete #5. 4/5

PR #256, opened 2017-04-04
Response to diagnostic

joebelmonte:


lib/diagnostic.js
```diff
@@ -32,14 +46,22 @@ const newArray = function newArray (/* length, defaultValue */) {
...
+
 // Questions 5
 // Write a function that adds or replaces a method on an object
 // using the given propertyName
 // If the argument obj is unset, create a new object
 // The function you write should return
 // the object passed in as an argument or the object created
 const addMethod = function addMethod (/* propertyName, method, obj */) {
-
+  obj[method] = propertyName
```
benjimelito, line 64: Yup, but don't forget to check if an object was passed in first. And also, you want to return the object at the end!

benjimelito:
Nice work, 4/5

PR #257, opened 2017-04-04
Diagnostic response

Pouk91:


lib/diagnostic.js
```diff
@@ -7,38 +7,56 @@
 // Write a function that takes zero or more arguments
 // and returns the minimum from that set
 // assume that all the argument are numbers
-const min = function min () {
-
+const min = function min (num1, num2) {
+  return num1 - num2
```
benjimelito, line 11: Yup, but you want this function to be able to take an unlimited number of arguments. You can use the `arguments` object inside your function

lib/diagnostic.js
```diff
@@ -7,38 +7,56 @@
...
+subtract(2, 1)
 
 // Question 2
 // Write a function that takes zero or more arguments
 // and returns the sum of that set
 // assume that all the argument are numbers
-const sum = function sum () {
-
+const sum = function sum (num1, num2) {
+  return num1 + num2
```
benjimelito, line 20: Same as above ^^

benjimelito:
Looks like you had a few issues with this one. No worries, we'll cover some of this during workshop later today. 0/5

PR #260, opened 2017-04-04
Complete js functions ins and outs diagnostic.

xiangcatherine:
Could get through questions 1-4, but didn't complete question 5.


benjimelito:
Looks good through question 4, I think if you had had a little more time you would have gotten them all. 4/5

PR #262, opened 2017-04-04
Response

mlynnes:
Attemp at completing diagnostic questions


benjimelito:
First two answers look good 👍  Please try to get these in before 9:30am though! (9:31am)

PR #263, opened 2017-04-04
Did as much as I could

d00medman:
Had major technical issues for the first 20-25 min of this diagnostic (terminal would not open, forcing me to hard reset my computer). Was able to get this much done.


benjimelito:
Actually looks like you got a lot done all things considered. Hope your terminal issues are cleared up at this point. 3/5

